PR #8, opened 2019-08-24
feat: display all assets on a map

e00dan:
screenshot-localhost_3000-2019 08 25-00_50_44

- Add AssetMap component and use it to display Production Assets map
- If no Origin contract lookup address look it up in backend
- Move configuration initialization from AppContainer to sagas
- Use dedicated file to handle generating routing URLs
- Create api.ts for storing methods for calling API
- Add Router to redux state
- Get component props from Redux state rather than passing through multiple layers from parent
- Add search test for ProducingAssetTable


kosecki123 (approved):
Good job 💯

PR #10, opened 2019-08-26
Node requirements set to 10. Typescript target set to es2018.

kosecki123:
- ts target set to es2018 which is fully supported by all node 10 versions https://node.green/#ES2018


e00dan (approved):
Love this!

PR #18, opened 2019-09-04
feat: Event Listener

josipbagaric:
# Energy Web Foundation: Event Listener

### Configuration

Create a `.env.dev` file in the root of the package with the following contents:

**.env.dev**
```
MODE="development"
UI_BASE_URL="http://localhost:3000"
API_BASE_URL="http://localhost:3030"
WEB3="http://localhost:8545"
MANDRILL_API_KEY="REPLACE_WITH_YOUR_API_KEY"
EMAIL_FROM="origin-no-reply@energyweb.org"
EMAIL_REPLY_TO="reply-to@energyweb.org"
```

### Run

From the root of the monorepo, run the following commands in two separate terminals:
1. `yarn run:origin`
2. `yarn run:event-listener`

The listener is now running and will react to every event that might occur on the blockchain.

## About 

Event listeners listen to events on the blockchain and react accordingly.

Components of this repo:
1. **Event Service**
    - Contains and manages starting/stopping multiple listeners
2. **Listener**
    - A listener is a custom component that 
3. **Email service and adapters**
    - Adapters for different email services


packages/event-listener/package.json
```diff
@@ -0,0 +1,70 @@
...
+        "mandrill-api": "1.0.45",
+        "nodemailer": "6.3.0",
+        "web3": "1.0.0-beta.37"
+    },
+    "devDependencies": {
+        "@types/web3": "1.0.14",
+        "axios": "0.19.0",
+        "dotenv": "8.0.0",
+        "ganache-cli": "6.6.0",
+        "lint-staged": "8.1.7",
```
kosecki123, line 51: Unless you need specific version, all those tools are installed in top level package.json. You can remove the devDeps from here.

josipbagaric: I only removed the ones that are in the top package.json. These are all the ones that aren't.

e00dan (approved):

PR #28, opened 2019-09-09
Notifications - Supply for demand

josipbagaric:
Screenshot 2019-09-10 at 14 45 36

- [x] Extends the event listener to listen for `LogPublishForSale` events and notify demand owners of new supply found for a demand
- [x] Introduces a flag `notifications` to the User, which can be toggled to turn email notifications ON/OFF
- [x] Change the approach to testing in the event-listener
- [x] Show the users their email in the UI
- [x] Add a notification toggle to the UI
- [x] Fix `EventHandlerManager` handling the same block when using ganache without blocktime


packages/utils-demo/package.json
```diff
@@ -12,7 +12,7 @@
     "build-ts": "tsc -b tsconfig.build.json --verbose --pretty",
     "deploy-empty": "yarn build-ts && node dist/js/deployEmpty",
     "start": "yarn build-ts && node dist/js/src/test",
-    "start-all": "concurrently --restart-tries 3 --restart-after 1500 -n eth,backend,deploy \"yarn start-ganache\" \"yarn start-test-backend\" \"yarn start\"",
+    "start-all": "concurrently --restart-tries 3 --restart-after 1500 -n eth,backend,deploy \"yarn start-ganache\" \"yarn start-test-backend\" \"sleep 10 && yarn start\"",
```
e00dan: Use `\"wait-on tcp:8545 && yarn start\"` instead of `sleep`

packages/event-listener/package.json
```diff
@@ -23,7 +23,7 @@
         "test": "rm -rf db.json && npm run build && mocha dist/js/test/ --timeout 60000 --exit",
         "test:concurrent": "concurrently --success first --kill-others -n eth,backend,test \"yarn start-ganache\" \"yarn start-test-backend\" \"yarn test\"",
         "start-test-backend": "node ../../node_modules/@energyweb/utils-testbackend/dist/js/src/index.js",
-        "start-ganache": "ganache-cli -m 'chalk park staff buzz chair purchase wise oak receive avoid avoid home' -l 8000000 -e 1000000 -a 20",
+        "start-ganache": "ganache-cli -m 'chalk park staff buzz chair purchase wise oak receive avoid avoid home' -l 8000000 -e 1000000 -a 20 -b 0,5",
```
e00dan: Is `-b` flag required?

It's discouraged by Ganache CLI:
```
Using the --blockTime flag is discouraged unless you have tests which require a specific mining interval.
```

josipbagaric: It is needed in this case to make the tests work, otherwise it doesn't handle events properly (if there is an event in the last block, it will keep sending notifications until there is a new block mined - which is never).

packages/utils-demo/package.json
josipbagaric: Thanks. These are some leftovers from my testing, not related to this PR (still WIP).

packages/event-listener/package.json
josipbagaric: Will work on a better solution once the features are done.

packages/origin-ui/src/components/Header.tsx
```diff
@@ -19,23 +19,38 @@ import { NavLink, withRouter, RouteComponentProps } from 'react-router-dom';
...
 import './Header.scss';
 import { connect } from 'react-redux';
 import { IStoreState } from '../types';
-import { getBaseURL, getCurrentUser } from '../features/selectors';
+import { getBaseURL, getCurrentUser, getConfiguration } from '../features/selectors';
 import { getAssetsLink, getCertificatesLink, getDemandsLink } from '../utils/routing';
+import { AccountDetailsModal } from '../elements/Modal/AccountDetailsModal';
 
-interface IStateProps {
+interface IHeaderProps {
```
e00dan: Could we keep the convention to name state properties (properties coming from Redux state) `IStateProps`?

If component has its own props then IMO it should be called `IOwnProps`.

`IHeaderProps` is all of the properties, which in this case would be:
```
type Props = RouteComponentProps & IStateProps;
```

packages/origin-ui/src/elements/Modal/AccountDetailsModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+        this.saveChanges = this.saveChanges.bind(this);
+        this.toggleNotifications = this.toggleNotifications.bind(this);
+
+        this.state = {
+            show: props.showModal,
+            notificationsEnabled: props.currentUser.offChainProperties.notifications
+        };
+    }
+
+    UNSAFE_componentWillReceiveProps(newProps: IAccountDetailsModalProps) {
```
e00dan: We should stop using this hook

packages/utils-general/src/blockchain-facade/ContractEventHandler.ts
```diff
@@ -34,14 +35,16 @@ export class ContractEventHandler {
...
-        const events = await this.contractInstance.getWeb3Contract().getPastEvents('allEvents', {
-            fromBlock: Math.min(this.lastBlockChecked + 1, blockNumber),
-            toBlock: blockNumber
-        });
-        this.unhandledEvents = events.reverse().concat(this.unhandledEvents);
-        this.lastBlockChecked =
-            blockNumber > this.lastBlockChecked ? blockNumber : this.lastBlockChecked;
-        this.walkThroughUnhandledEvent();
+
+        if (blockNumber > this.lastBlockChecked) {
```
e00dan, line 39: :clap: :heart_eyes:

e00dan (approved):
Great job, minor feedback

kosecki123 (approved):

PR #31, opened 2019-09-10
chore(market-matcher): market matcher next

kosecki123:
A re-designed version of the `market-matcher`

- no magic json configuration files
- no scheme generation - just JSON validation
- code refactoring
- stale simulator removal
- code strategies support
- uses dotenv configs


packages/market-matcher/scripts/buildSchemas.js
```diff
@@ -76,8 +76,6 @@ async function run() {
 
     await generateSchemaIfNew('yarn --silent generate-rule-schema', '/schemas/rule.schema.json');
 
-    await generateSchemaIfNew('yarn --silent generate-conf-schema', '/schemas/conf.schema.json');
```
e00dan, line 79: May God bless for each schema less

packages/market-matcher/src/EntityStore.ts
```diff
@@ -0,0 +1,207 @@
...
+        this.logger.verbose('Registered new supply #' + supply.id);
+    }
+
+    private registerAgreement(agreement: Agreement.Entity) {
+        const allowed = agreement.allowedMatcher.some(
+            matcherAddress => matcherAddress && matcherAddress.toLowerCase() === this.matcherAddress
+        );
+
+        if (!allowed) {
+            this.logger.verbose('This instance is not an matcher for agreement #' + agreement.id);
```
e00dan, line 197: `an matcher` => `a matcher`

e00dan (approved):
Impressive, great work!

packages/market-matcher/.env.example
```diff
@@ -0,0 +1,6 @@
+WEB3 = "http://localhost:8545"
+MARKET_CONTRACT_ADDRESS = "0x35E1bA1dEC0bc5A14799563EF8564310846D1438"
+ORIGIN_CONTRACT_ADDRESS = "0xE377ae7241f76aC5Cdb89f5DEbc0dBE82f9c2C03"
+MATCHER_PUB_KEY = "0x585cc5c7829b1fd303ef5c019ed23815a205a59e"
```
josipbagaric, line 4: Do we need to have the public key defined? Can it not be derived from the private key?

packages/market-matcher/package.json
```diff
@@ -14,20 +14,15 @@
     "version": "1.1.0",
     "main": "dist/js/src/index.js",
     "scripts": {
-        "build": "yarn generate-schemas && yarn build-ts",
+        "build": "yarn build-ts",
```
josipbagaric: Can we just replace `build-ts` with `build`?

josipbagaric (approved):
Looks good. Would be cool to enforce types, as a lot of variables are untyped (copy-paste of old code, I assume).

packages/market-matcher/package.json
kosecki123: So build should clean and build-ts actually, which is here omitted.

packages/market-matcher/.env.example
kosecki123: Yeah, good point.

kosecki123:
> Looks good. Would be cool to enforce types, as a lot of variables are untyped (copy-paste of old code, I assume).

Thanks! `noImplicitAny` enabled for the `market-matcher` project now.

PR #36, opened 2019-09-13
Simplified matcher checking logic + fillDemand action

josipbagaric:
- Remove storage references to matcher
- Allow any user with the role `Matcher` to perform matching
- Add a `fillDemand` action to MarketLogic
- Add event listener for `fillDemand`
- Remove storing matcher properties off-chain together with a agreement properties


e00dan (approved):

packages/market/contracts/Trading/MarketLogic.sol
```diff
@@ -89,6 +95,31 @@ contract MarketLogic is AgreementLogic {
...
+    /// @param _entityId ID of the tradable entity
+    function fillDemand(
+        uint _demandId,
+        uint _entityId
+    )
+        external
+        onlyRole(RoleManagement.Role.Matcher)
+    {
+        MarketDB.Demand memory demand = db.getDemand(_demandId);
+        require(demand.status != MarketDB.DemandStatus.ARCHIVED, "demand cannot be in archived state");
```
kosecki123: Should be only in ACTIVE state. PAUSED is illegal here too.

packages/market-matcher/src/MatchableAgreement.ts
```diff
@@ -25,19 +23,9 @@ export class MatchableAgreement {
             .result();
     }
 
-    public async missingEnergyForDemand(demand: Demand.IDemand, config: Configuration.Entity) {
-        const { start, timeframe } = this.agreement.offChainProperties;
-        const currentPeriod = await Utils.getCurrentPeriod(start, timeframe, config);
-
+    public async missingEnergyForDemand(demand: Demand.IDemand) {
         const { targetWhPerPeriod } = demand.offChainProperties;
```
kosecki123, line 27: Are we tracking the missing energy using `targetWhPerPeriod` ?

josipbagaric: Tbh I just removed this because I was removing separate references to matchers for every agreement - thought we will be refactoring later in the sprint.

packages/market/contracts/Trading/MarketLogic.sol
```diff
@@ -1,39 +1,29 @@
...
 import "../../contracts/Trading/AgreementLogic.sol";
-import "@energyweb/asset-registry/contracts/Interfaces/AssetGeneralInterface.sol";
 
 /// @title The logic contract for the AgreementDB of Origin list
 contract MarketLogic is AgreementLogic {
     event createdNewDemand(address _sender, uint indexed _demandId);
     event createdNewSupply(address _sender, uint indexed _supplyId);
     event DemandStatusChanged(address _sender, uint indexed _demandId, uint16 indexed _status);
     event DemandUpdated(uint indexed _demandId);
+    event DemandFilled(uint indexed _demandId, uint indexed _entityId);
```
kosecki123: What is the means of this event? Fully filled? And if yes, what about partial fills? I think we should support demand being filled by multiple certificates.

packages/market-matcher/src/MatchableAgreement.ts
josipbagaric: And I guess this build passed because there is no test written that checks if the demand size decreases. This should be fixed.

packages/market/contracts/Trading/MarketLogic.sol
josipbagaric: Probably the naming is not clear enough. This is just an event that says that a certificate has been matched to a demand (partially filled).

kosecki123: Can we re-construct using this event the amount of energy that was matched during matching event?

packages/event-listener/src/listeners/origin.listener.ts
```diff
@@ -51,7 +51,7 @@ export class OriginEventListener implements IOriginEventListener {
         this.originLookupAddress = originLookupAddress;
         this.web3 = web3;
         this.emailService = emailService;
-        this.notificationInterval = notificationInterval || 60000; // Default to 1 min intervals
+        this.notificationInterval = notificationInterval || 6000; // Default to 1 min intervals
```
e00dan: This isn't 1 min interval default anymore, right? (comment)

packages/event-listener/src/listeners/origin.listener.ts
```diff
@@ -89,6 +94,15 @@ export class OriginEventListener implements IOriginEventListener {
             await this.checkDemands(publishedCertificate);
         });
 
+        marketContractEventHandler.onEvent('DemandFilled', async (event: any) => {
+            const { _demandId, _entityId } = event.returnValues;
+            // const transferredCert = await new Certificate.Entity(_entityId, this.conf).sync();
```
e00dan, line 99: Remove commented-out code

packages/market-matcher/src/test/Matcher.test.ts
```diff
@@ -170,11 +170,27 @@ describe('Test StrategyBasedMatcher', async () => {
...
             privateKeyDeployment
         );
-        marketLogic = new MarketLogic(web3, (deployedContracts as any).MarketLogic);
+
+        const marketLogicAddress: string = (deployedContracts as any).MarketLogic;
+
+        marketLogic = new MarketLogic(web3, marketLogicAddress);
         marketContractLookupAddr = (deployedContracts as any).MarketContractLookup;
 
+        await userLogic.createUser(
```
e00dan: We should leverage User.createUser where possible

e00dan (approved):

kosecki123 (approved):

PR #40, opened 2019-09-17
fix: Handle onSale certificates. 

kosecki123:
- properly handle `onSale` flag
- listen to `LogPublishOnSale` event
- process events in a sequential fashion
- emit `LogUnpublishOnSale` after successful buy

image

### DIFF
--- a/packages/market-matcher/src/CertificateService.ts
+++ b/packages/market-matcher/src/CertificateService.ts
@@ -16,9 +16,14 @@ export class CertificateService {
 
     public async matchAgreement(certificate: Certificate.Entity, agreement: Agreement.IAgreement) {
         const demand = this.entityStore.getDemandById(agreement.demandId.toString());
+        if (await this.isAlreadyTransferred(certificate, demand.demandOwner)) {
+            return;
+        }
+
         this.logger.debug(
             `Transferring certificate to ${demand.demandOwner} with account ${this.config.blockchainProperties.activeUser.address}`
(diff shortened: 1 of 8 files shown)


packages/market-matcher/src/CertificateService.ts
```diff
@@ -49,10 +54,31 @@ export class CertificateService {
...
     }
+
+    private async isAlreadyTransferred(certificate: Certificate.Entity, owner: string) {
+        const syncedCertificate = await certificate.sync();
+
+        this.logger.verbose(
+            `isAlreadyTransferred: #${syncedCertificate.id} owned by ${syncedCertificate.owner}`
+        );
+
+        if (certificate.owner === owner) {
```
e00dan: I recently fixed bunch of problems where addresses were the same, but different case. Is toLowerCase needed here?

packages/market-matcher/src/Matcher.ts
```diff
@@ -82,6 +101,11 @@ export class Matcher {
     }
 
     private async matchWithDemands(certificate: Certificate.Entity) {
+        if (!this.isOnSale(certificate)) {
+            this.logger.verbose(`This certificate is on for sale #${certificate.id}`);
```
e00dan: Should this be `This certificate is not for sale`?

e00dan (approved):

josipbagaric (approved):

PR #44, opened 2019-09-19
feat(origin-ui): filter, search claimed certificates, bulk claim in inbox

e00dan:
- add filter and search to claimed certificates
- add bulk claiming certificates
- allow strictly typed blockchain configuration across a package
- give more time for matcher tests to complete to avoid breaking tests on master
- remove package-lock.json from market and update gitignore


packages/origin/contracts/Origin/CertificateLogic.sol
```diff
@@ -158,6 +158,39 @@ contract CertificateLogic is CertificateInterface, CertificateSpecificContract,
...
+
+    function retireCertificateInternal(uint _certificateId, address claimer) internal {
+        CertificateDB.Certificate memory cert = CertificateDB(address(db)).getCertificate(_certificateId);
+        require(cert.tradableEntity.owner == claimer, "You have to be the owner of the contract.");
+        require(
+            cert.certificateSpecific.children.length == 0,
+            "Unable to retire certificates that have already been split."
+        );
+
+        if (cert.certificateSpecific.status != uint(CertificateSpecificContract.Status.Retired)) {
```
josipbagaric: Would it make more sense to replace this with `require(cert.certificateSpecific.status != uint(CertificateSpecificContract.Status.Retired), "cannot retired a certificate that has already been retired");`? So that it doesn't silently fail.

packages/origin/contracts/Origin/CertificateLogic.sol
```diff
@@ -158,6 +158,39 @@ contract CertificateLogic is CertificateInterface, CertificateSpecificContract,
         }
     }
 
+    /// @notice claims a set of certificates
+    /// @param _idArray the ids of the certificates to be claimed
+    function claimCertificateBulk(uint[] calldata _idArray) external {
```
josipbagaric, line 160: I would limit this to a certain maximum array size like `uint[10] calldata _idArray` or something, to stop overflows on block gas limit.

packages/origin/contracts/Origin/CertificateLogic.sol
```diff
@@ -158,6 +158,39 @@ contract CertificateLogic is CertificateInterface, CertificateSpecificContract,
...
+        for (uint i = 0; i < _idArray.length; i++) {
```
kosecki123, line 162: This potentially can run out of gas, UI seems to limit to 100 items per call. I would check the upper limit on this.

packages/origin/src/blockchain-facade/Certificate.ts
```diff
@@ -388,3 +375,17 @@ export class Entity extends TradableEntity.Entity implements ICertificate {
         return returnEvents;
     }
 }
+
+export async function claimCertificates(
```
josipbagaric, line 390: Would be cool if we would also add `await certificate.claim()` method to Certificate.ts. This would allow claiming a certificate without passing in an array of IDs.

josipbagaric:
Also @Kuzirashi, would be cool to write tests for bulk claiming.

e00dan:
It seems that claiming certificates could go maybe up to 180, but I decided to put a limit on 100, because:
1. we don't a have a requirement right now to go above that
2. safety margin for errors / changes / increasing gas costs of claiming etc.
3. we already have a limit of 100 on bulk buying

josipbagaric (approved):

packages/origin/src/blockchain-facade/Certificate.ts
```diff
@@ -371,3 +374,29 @@ export class Entity extends TradableEntity.Entity implements ICertificate {
...
+
+    const accountProperties = {
+        from: configuration.blockchainProperties.activeUser.address,
+        privateKey: configuration.blockchainProperties.activeUser.privateKey
+    };
+
+    return configuration.blockchainProperties.certificateLogicInstance.claimCertificateBulk(certificateIdsAsNumber, accountProperties);
+};
+
+export async function claim(
```
josipbagaric: Would be better to make this a part of `Certificate.Entity`. Wouldn't have to pass the ID every time then, but you could just call `await certificate.claim()`.

PR #82, opened 2019-09-23
feat: demand fill tracking

kosecki123:
This PR provides the methods to calculate missing energy in demand in total and per period. `market-matcher` is now skipping demand with the current period already filled. 

The solution is based on `DemandPartiallyFilled` event from the `market`, uses it to aggregate information about filled demand periods. Event emit timestamp is assumed to be the time of fill operation.

This also fixes the bug where only certificates with higher or matching energy were matched with the demand.

Untitled Diagram


packages/market/src/blockchain-facade/Demand.ts
```diff
@@ -229,3 +250,108 @@ export const deleteDemand = async (
...
+};
+
+export const calculateMissingEnergyDemand = async (
+    demand: IDemand,
+    config: Configuration.Entity
+) => {
+    const { startTime, endTime, timeFrame, energyPerTimeFrame } = demand.offChainProperties;
+
+    if (timeFrame === TimeFrame.hourly || timeFrame === TimeFrame.halfHourly) {
+        throw new Error('Hourly and half-hourly demands are not supported');
```
josipbagaric: Why not?

josipbagaric (approved):

packages/asset-registry/src/blockchain-facade/ConsumingAsset.ts
```diff
@@ -72,7 +72,7 @@ export const getAllAssetsOwnedBy = async (owner: string, configuration: Configur
 
 export class Entity extends Asset.Entity implements Asset.IOnChainProperties {
     getUrl(): string {
-        const consumingAssetLogicAddress = this.configuration.blockchainProperties.consumingAssetLogicInstance.web3Contract._address;
+        const consumingAssetLogicAddress = this.configuration.blockchainProperties.consumingAssetLogicInstance.web3Contract.options.address;
```
e00dan, line 75: Is this preferred way of accessing address?

packages/event-listener/src/listeners/origin.listener.ts
```diff
@@ -204,8 +204,8 @@ export class OriginEventListener implements IOriginEventListener {
             this.conf
         ).sync();
 
-        const matchedDemands = demands.filter(demand => {
-            const { result } = new MatchableDemand(demand).matchesCertificate(
+        const matchedDemands = demands.filter(async demand => {
+            const { result } = await new MatchableDemand(demand).matchesCertificate(
```
e00dan: Is this construct going to work? `await new` looks unfamiliar to me

packages/origin-ui/src/__tests__/DemandForm.test.tsx
```diff
@@ -210,13 +218,13 @@ describe('DemandForm', () => {
         expect(submitButton.getDOMNode().hasAttribute('disabled')).toBe(false);
 
         createDemand = (offChainProps: Demand.IDemandOffChainProperties) => {
-            expect(moment.unix(parseInt(offChainProps.startTime, 10)).format('YYYY-MM-D')).toBe(
+            expect(moment.unix(offChainProps.startTime).format('YYYY-MM-D')).toBe(
```
e00dan, line 221: :clap:

packages/origin-ui/src/components/Demand/DemandForm.tsx
```diff
@@ -41,34 +41,6 @@ import { FormikDatePicker } from '../FormikDatePicker';
 import { AssetTypeSelector } from '../AssetTypeSelector';
 import { getCurrentUser } from '../../features/users/selectors';
 
-export function calculateTotalEnergyDemand(
```
e00dan, line 44: :clap:

e00dan (approved):
:+1:

packages/asset-registry/src/blockchain-facade/ConsumingAsset.ts
kosecki123: Yeah. That seems to be the only way if you follow the `Contract` type. which we now do.

packages/event-listener/src/listeners/origin.listener.ts
kosecki123: `new` has higher precedence than `await` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#Table

PR #85, opened 2019-09-23
chore(origin-ui): add tests for demand form

e00dan:
- add tests for demand form
- enable eslint with slightly modified configuration for ui tests

### DIFF
--- a/packages/origin-ui/src/__tests__/DemandForm.test.tsx
+++ b/packages/origin-ui/src/__tests__/DemandForm.test.tsx
@@ -1,0 +1,250 @@
+import * as React from 'react';
+import { mount } from 'enzyme';
+import { Provider } from 'react-redux';
+import { createStore, applyMiddleware } from 'redux';
+import { createMemoryHistory } from 'history';
+import { createRootReducer } from '../reducers';
+import { dataTestSelector } from '../utils/Helper';
+import createSagaMiddleware from 'redux-saga';
+import { routerMiddleware, ConnectedRouter } from 'connected-react-router';
+import sagas from '../features/sagas';
+import { DemandForm } from '../components/Demand/DemandForm';
+import { MuiPickersUtilsProvider } from '@material-ui/pickers';
+import MomentUtils from '@date-io/moment';
+import { TimeFrame, Currency } from '@energyweb/utils-general';
+import { addUser, updateCurrentUserId } from '../features/users/actions';
+import { User } from '@energyweb/user-registry';
+import { Demand } from '@energyweb/market';
+import moment from 'moment';
+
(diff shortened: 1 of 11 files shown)


packages/origin-ui/src/__tests__/DemandForm.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+            delete offChainProps.startTime;
+            delete offChainProps.endTime;
+
+            expect(offChainProps).toStrictEqual({
+                currency: Currency.EUR,
+                timeFrame: TimeFrame.daily,
+                maxPricePerMwh: 100,
+                targetWhPerPeriod: 1000000
+            });
```
kosecki123, line 234: Maybe use Partial<Demand.IDemandOffChainProperties> instead of deleting startTime and endTime.

packages/origin-ui/src/__tests__/DemandForm.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+                    .set('date', dayOfMonth)
+                    .format('MMMM Do')
+            );
+
+            // Close Datepicker (click outside)
+            (document.querySelector('body > [role="presentation"] > div') as HTMLElement).click();
+
+            await refresh();
+
+            await waitForConditionAndAssert(
```
kosecki123, line 138: Can we use some oldschool js events like onLoad?

kosecki123 (approved):

packages/origin-ui/src/__tests__/DemandForm.test.tsx
e00dan: Marking it as Partial wouldn't delete these properties - which is needed because strict comparing is performed between the arrays.

packages/origin-ui/src/__tests__/DemandForm.test.tsx
e00dan: Maybe it would work, there's MutationObserver that could trigger events, I'm not sure if js-dom and enzyme support it. I'll experiment with it in future.

packages/origin-ui/src/__tests__/DemandForm.test.tsx
kosecki123: Check this out https://github.com/energywebfoundation/origin/blob/50b4b8074e473c09636392b5b492df63c4b30ea3/packages/market/src/test/MarketFacade.test.ts#L265

e00dan: Aah, didn't know such function exists. Thanks @kosecki123

PR #87, opened 2019-09-25
Chore: Turn on noImplicitAny

josipbagaric:
Enable `noImplicitAny` rule globally, disable only in `origin-ui` for now.

### DIFF
--- a/packages/asset-registry/src/wrappedContracts/AssetConsumingDB.ts
+++ b/packages/asset-registry/src/wrappedContracts/AssetConsumingDB.ts
@@ -4,24 +4,24 @@ import AssetConsumingDBJSON from '../../build/contracts/lightweight/AssetConsumi
 
 export class AssetConsumingDB extends GeneralFunctions {
     web3: Web3;
-    buildFile = AssetConsumingDBJSON;
 
     constructor(web3: Web3, address?: string) {
+        const buildFile: any = AssetConsumingDBJSON;
         super(
             address
-                ? new web3.eth.Contract(AssetConsumingDBJSON.abi, address)
+                ? new web3.eth.Contract(buildFile.abi, address)
                 : new web3.eth.Contract(
-                      AssetConsumingDBJSON.abi,
-                      (AssetConsumingDBJSON as any).networks.length > 0
-                          ? AssetConsumingDBJSON.networks[0]
-                          : null
-                  )
+                    buildFile.abi,
+                    buildFile.networks.length > 0
+                        ? buildFile.networks[0]
+                        : null
+                )
         );
         this.web3 = web3;
     }
 
     async getAllLogChangeOwnerEvents(eventFilter?: SearchLog) {
-        let filterParams;
+        let filterParams: any;
         if (eventFilter) {
             filterParams = {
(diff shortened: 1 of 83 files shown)


packages/asset-registry/src/wrappedContracts/AssetConsumingDB.ts
```diff
@@ -4,24 +4,24 @@ import AssetConsumingDBJSON from '../../build/contracts/lightweight/AssetConsumi
 
 export class AssetConsumingDB extends GeneralFunctions {
     web3: Web3;
-    buildFile = AssetConsumingDBJSON;
 
     constructor(web3: Web3, address?: string) {
+        const buildFile: any = AssetConsumingDBJSON;
```
e00dan, line 9: This single change looks like a step backward.

Using :any here drops the types that should be imported from JSON (as far as I remember TS correctly resolved them when I was refactoring JSON's to reduce build size).

packages/asset-registry/src/wrappedContracts/AssetConsumingRegistryLogic.ts
```diff
@@ -5,24 +5,24 @@ import moment from 'moment';
 
 export class AssetConsumingRegistryLogic extends GeneralFunctions {
     web3: Web3;
-    buildFile = AssetConsumingRegistryLogicJSON;
 
     constructor(web3: Web3, address?: string) {
+        const buildFile: any = AssetConsumingRegistryLogicJSON;
```
e00dan, line 10: Same comment, we're dropping the types resolved from JSON by TS here by using :any

packages/asset-registry/src/wrappedContracts/AssetConsumingDB.ts
e00dan: If there's a problem with types maybe you can use:

`buildFile: typeof AssetConsumingDBJSON` or `buildFile: AssetConsumingDBJSON`

e00dan (approved):
Step in a good direction :+1: 

Can we move `noImplicitAny: true` rule to global tsconfig and disable it only for UI?

This way when someone starts a new project they inherit from our standards :)

josipbagaric:
> Step in a good direction 👍
> 
> Can we move `noImplicitAny: true` rule to global tsconfig and disable it only for UI?
> 
> This way when someone starts a new project they inherit from our standards :)

That's what I've done, as mentioned in the PR description.

packages/asset-registry/src/wrappedContracts/AssetConsumingDB.ts
josipbagaric: All of those cause type issues. We could look for a better solution, but it's probably not worth the time/effort, assuming wrapped contracts will be removed/replaced soon enough.

packages/asset-registry/src/wrappedContracts/AssetConsumingRegistryLogic.ts
josipbagaric: Those seem to not be resolved properly since we moved away from `"target": "commonjs"` in tsconfig.

kosecki123 (approved):

PR #89, opened 2019-09-26
feat: Use I-REC asset types across system

e00dan:
- Multi-level asset type certificates filtering in UI
- Use more comprehensive implementation of asset types overlap checking
- Fix UI bug where there would be an error in the console if MetaMask account is not registered as user
- Adjust demo-config to work with new types
- Add deploying demo to Travis CI


packages/utils-general/src/blockchain-facade/AssetTypeService.ts
```diff
@@ -88,7 +88,48 @@ export class IRECAssetService implements IAssetService {
...
+                    assetType.every(
+                        (type, index) =>
+                            type === decodedTypeToCheck[index] || !decodedTypeToCheck[index]
+                    ))
+            ) {
+                return true;
+            }
+        }
+
+        return false;
```
kosecki123: This looks suspiciously complex, why do you need that? Please add the unit test with input that requires those changes.

e00dan: Yep, it's in progress. I need this because:

If only wind (level 1) is selected it should match with wind;onshore (level 2)

same with level 3 being matched with level 2, unless level 3 specificity is present for that concrete category

kosecki123: Yes, this is by design. This method is used by matcher to check for compatible asset types. Now if the demand is broad and requests `Wind` then whatever level 2 or level 3 type of `Wind` is defined by `ProducingAsset`, it matches.

Take a look at tests here: https://github.com/energywebfoundation/origin/blob/5a93304f837c317f6940f35ee37113fef66616eb/packages/utils-general/src/test/AssetTypeService.test.ts#L71

kosecki123: So instead of changing the impl of this method please add the new one for the purpose.

kosecki123 (changes requested):

packages/utils-general/src/blockchain-facade/AssetTypeService.ts
e00dan: The test is there but it was failing, you can check on master. Also I believe, it doesn't check correctly when `Wind` and `Wind;Onshore` for Wind:Offshore would be in the same array, because it would match `Wind;Offshore` with `Wind`, while it shouldn't match, because there's a level 2 specificity passed - `Wind;Onshore`.

kosecki123: Yes, you are right, test was failing. #90 fixes it.

josipbagaric (approved):

kosecki123 (approved):

PR #107, opened 2019-09-29
chore(origin-ui): add certificate table test

e00dan:
- Add Certificate Table test
- Remove `.only` from Matcher test case
- Organize parts of UI test code into reusable helpers

### DIFF
--- a/packages/origin-ui/src/__tests__/utils/demo.ts
+++ b/packages/origin-ui/src/__tests__/utils/demo.ts
@@ -1,0 +1,225 @@
+import Web3 from 'web3';
+import ganache from 'ganache-cli';
+import * as Winston from 'winston';
+import { migrateUserRegistryContracts } from '@energyweb/user-registry/contracts';
+import { migrateAssetRegistryContracts } from '@energyweb/asset-registry/contracts';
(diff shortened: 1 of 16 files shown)


josipbagaric (approved):

packages/origin-ui/src/__tests__/utils/demo.ts
```diff
@@ -0,0 +1,225 @@
...
+            marketLogicInstance: marketLogic,
+            web3
+        },
+        offChainDataSource: {
+            baseUrl: API_BASE_URL
+        },
+        logger
+    };
+
+    const adminPropsOnChain: User.IUserOnChainProperties = {
```
kosecki123, line 129: I suggest taking a look at Substitute mocking lib used for e.g in `market-matcher` for faster test objects generation https://github.com/energywebfoundation/origin/blob/2948e17dbba72b6e74e737b6a8b811b2234fbb42/packages/market-matcher/src/test/unit/MatchableAgreement.test.ts#L29

You would avoid creating those full-fledged objects yet with empty properties.

kosecki123 (approved):

PR #126, opened 2019-10-10
feat(origin-ui): using in-browser private key

e00dan:
- Add option to import private key and store it in browser encrypted with a password
- Provide option to unlock previously saved private key and use it
- Allow browsing site without MetaMask by fallback to default web3 RPC
- Start using hooks (which simplify whole component structure, for example - no need to declaring types as much as it was before)
- Add settings icon next to user indicator to navigate to Account settings
- Make demand creation test end-to-end, so it tests also blockchain layer
- Use WebSocket RPC ws://localhost:8545 in tests, because http one fails for some reason
- Update JSdom in tests from v11 to v15
- Update material UI, closes https://github.com/energywebfoundation/origin/pull/122

localhost_3000_account_import
localhost_3000_account_import (1)
localhost_3000_account_import (2)
localhost_3000_account_import (3)
localhost_3000_account_import (4)


e00dan:
Closing in favor of https://github.com/energywebfoundation/origin/pull/136

PR #137, opened 2019-10-14
feat(backend): Replace testbackend with the new origin-backend

josipbagaric:
Replaced `utils-testbackend` with a new `origin-backend`.

Features:
- New ORM structure with the following entities:
  - `MarketContractLookup` - contains a reference to all the Market contracts deployed
  - `AnyEntity` - serves as a multi-functional key-value store-it-all entity
- Use of [TypeORM](https://github.com/typeorm/typeorm/) for easier adaptation to different databases
- Store test databases in SQLite instead of a JSON file
- Add an API versioning scheme: `HOST/api` (latest API version) and `HOST/api/v1` point to the same API for now, but it's extendable in the future
- Separate each route to a separate file and use a router to handle routes
- Better code structure and quality


kosecki123:
Since we reworking the backend I would like us to create `origin-backend-client` which consists of the code that will communicate with the API as a service as well as mock object that can be used for testing. This enables us to:
- get rid of direct calling API
- remove backend dependencies from the projects in favour of `origin-backend-client` 
- faster testing

packages/origin-backend/readme.md
```diff
@@ -0,0 +1,5 @@
+# Energy Web Foundation: origin-release C: origin-backend
```
Kuzirashi: I'm not sure if adding mention about `release C` to origin-backend readme adds valuable information to the document, maybe we can omit this information?

e00dan (approved):

packages/origin-backend/src/entity/AnyEntity.ts
```diff
@@ -0,0 +1,20 @@
+import { Entity, Column, BaseEntity, Unique, PrimaryGeneratedColumn } from 'typeorm';
+
+@Entity()
+@Unique(['type', 'contractAddress', 'identifier'])
+export class AnyEntity extends BaseEntity {
```
kosecki123: Why not naming it `Entity`?

PR #164, opened 2019-10-22
fix(event-listener): listen to process.env.MARKET_CONTRACT_ADDRESS

josipbagaric:
Simplify event listener by only listening to one contract at a time.

Uses `process.env.MARKET_CONTRACT_ADDRESS`. If not set, then uses the latest contract from the backend.


e00dan (approved):

packages/event-listener/src/index.ts
```diff
@@ -1,17 +1,37 @@
...
 
 export const SCAN_INTERVAL = 3000;
 
-const startEventListener = () => {
+const startEventListener = async () => {
     const web3 = new Web3(process.env.WEB3 || 'http://localhost:8550');
+    const backendUrl: string = process.env.BACKEND_URL || 'http://localhost:3035';
 
-    const eventService = new EventServiceProvider(process.env.BACKEND_URL, web3);
+    const result: AxiosResponse = await axios.get(`${backendUrl}/MarketContractLookup`);
```
kosecki123, line 16: When declaring the const you probably want to skip adding type explicitly. It's obviously correct as it it, perhaps we should agree on the style or/and enforce it via eslint.

kosecki123 (approved):

PR #191, opened 2019-10-26
fix: fix event-listener and simulator configuration

e00dan:
- solar-simulator: Get contract address from API
- event-listener: Fix getting latest contract from API (missing `/api`)
- market-matcher: Get contract address from API

New approach:
1. if specified, take Market address from .env
2. if not, get it from API

### DIFF
--- a/packages/solar-simulator/package.json
+++ b/packages/solar-simulator/package.json
@@ -19,6 +19,7 @@
   "license": "GPL-3.0-only",
   "dependencies": {
     "@energyweb/asset-registry": "1.3.0",
+    "@energyweb/market": "1.3.0",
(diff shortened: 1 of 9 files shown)


packages/solar-simulator/package.json
```diff
@@ -19,6 +19,7 @@
   "license": "GPL-3.0-only",
   "dependencies": {
     "@energyweb/asset-registry": "1.3.0",
+    "@energyweb/market": "1.3.0",
```
kosecki123, line 22: When adding dependency here please also add it to tsconfig as dependent project to keep the consistency.

kosecki123 (approved):

PR #193, opened 2019-10-27
fix(origin): fix getting all certificates events

e00dan:
- add fromBlock to getting events params (required for Volta RPC node to return events)


packages/origin-ui/src/components/CertificateDetailView.tsx
```diff
@@ -190,9 +190,10 @@ class CertificateDetailViewClass extends React.Component<Props, IDetailViewState
                     <span className="timestamp text-muted">
                         {new Date(event.timestamp * 1000).toLocaleString()} -{' '}
                         <a
-                            href={'https://tobalaba.etherscan.com/tx/' + event.txHash}
+                            href={'https://volta-explorer.energyweb.org/tx/' + event.txHash}
```
kosecki123: I suggest we move this URL to .env config.

kosecki123 (approved):

PR #259, opened 2019-11-13
feat(origin-ui): runtime Docker environment variables

e00dan:
Tested on:
- [x] Linux
- [x] OSX
- [x] Windows with Windows Subsystem for Linux 2

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,20 +1,20 @@
 # ------- REQUIRED --------- #
-MODE=production
+MODE="production"
 # URLs
-WEB3=http://localhost:8545
-BACKEND_URL=http://localhost:3030
+WEB3="http://localhost:8545"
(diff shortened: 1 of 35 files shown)


josipbagaric (approved):

.env.example
```diff
@@ -1,20 +1,20 @@
-# ------- NEEDED --------- #
-MODE="production"
+# ------- REQUIRED --------- #
```
kosecki123, line 1: There were some issues with unquoted values during docker builds. Don’t remember what exactly it was though.

e00dan: Okay, I see. I'll try to modify bash script to add `"` only if it's not present. Thanks for warning!

kosecki123:
Tested on OSX 🚀

kosecki123 (approved):

PR #260, opened 2019-11-14
PurchasableCertificate + MarketUser

josipbagaric:
- Separates all "purchasable" properties, such as price and currency from the bare `Certificate` and creates a separate Entity called `PurchasableCertificate` with purchasable properties. 
- Use `MarketUser` instead of regular `User` for marketplace specific settings


packages/market/src/blockchain-facade/PurchasableCertificate.ts
```diff
@@ -0,0 +1,394 @@
...
+    currency: Currency;
+}
+
+export const getCertificateListLength = async (
+    configuration: Configuration.Entity
+): Promise<number> => {
+    return Certificate.getCertificateListLength(configuration);
+};
+
+// const getAccountFromConfiguration = (configuration: Configuration.Entity) => ({
```
e00dan: Commented out code shouldn't be part of the commit

kosecki123 (approved):

e00dan (approved):

PR #293, opened 2019-11-19
feat: Add support for toggling between manual/automatic matcher

e00dan:


packages/market-matcher/src/EntityFetcher.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    getDemandListLength() {
+        return Demand.getDemandListLength(this.config);
+    }
+
+    getSupplyListLength() {
+        return Supply.getSupplyListLength(this.config);
+    }
+
+    getAgreement(id: string, tries = 1) {
```
kosecki123: Previously the default was 10. FYI 1 retry means 1 failed and 1 immediately after. This is our "hack" for awaiting db and blockchain be in sync. I recommend setting it to higher value, perhaps 5-10.

packages/market-matcher/src/test/unit/EntityStore.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        const config = Substitute.for<Configuration.Entity>();
+        const logger = Substitute.for<Winston.Logger>();
+        const fetcher = Substitute.for<IEntityFetcher>();
+
+        fetcher.getAgreementListLength().returns(Promise.resolve(0));
+        fetcher.getSupplyListLength().returns(Promise.resolve(0));
+        fetcher.getCertificateListLength().returns(Promise.resolve(0));
+        fetcher.getDemandListLength().returns(Promise.resolve(1));
+        fetcher.getDemand(Arg.all()).returns(Promise.resolve(demand));
```
kosecki123, line 32: Consider moving this block to a separate function, repeated 2 more times below.

kosecki123 (changes requested):
Please take a look at my comments.

packages/market-matcher/src/EntityFetcher.ts
e00dan: Set to 10

josipbagaric (approved):

kosecki123 (approved):

PR #295, opened 2019-11-19
chore: update web3 to 1.2.4

e00dan:


packages/origin/src/test/CertificateLogic.test.ts
```diff
@@ -546,7 +539,7 @@ describe('CertificateLogic-Facade', () => {
 
         const certificate = await new Certificate.Entity('6', conf).sync();
 
-        blockCreationTime = (await web3.eth.getBlock('latest')).timestamp;
+        blockCreationTime = parseInt((await web3.eth.getBlock('latest')).timestamp.toString(), 10);
```
kosecki123, line 542: What is the reason for parse(string), https://web3js.readthedocs.io/en/v1.2.1/web3-eth.html#getblock timestamp should be number already.

packages/origin/src/blockchain-facade/Certificate.ts
```diff
@@ -131,14 +133,21 @@ export const getAllCertificateEvents = async (
...
+
     public energy: number;
+
     public status: Status;
+
     public creationTime: number;
+
     public parentId: number;
+
     public children: string[];
```
kosecki123, line 151: Are these whitespace add manually or we somehow have different prettier configs?

kosecki123 (approved):

packages/origin/src/blockchain-facade/Certificate.ts
josipbagaric: Yeah this is added automatically, not a fan of this either. We should look into removing it IMO.

packages/user-registry/package.json
```diff
@@ -22,7 +22,7 @@
     "test": "test"
   },
   "scripts": {
-    "build": "yarn build:ts",
+    "build": "yarn build:static && yarn build:ts",
```
josipbagaric, line 25: Nice one. I think we should change `yarn build` in every repo to behave like this.

josipbagaric (approved):

packages/origin/src/test/CertificateLogic.test.ts
e00dan: Unfortunately docs aren't up to date.

typess

PR #297, opened 2019-11-20
chore: create market-matcher-core library

kosecki123:
This PR includes
- `market-matcher-core` library which holds matching logic, a new component in the origin SDK 
- `event-listener` and `origin-ui` uses `market-matcher-core` instead of CLI program
- removed `tsyringe` dependecy injection lib as it requires all injectable classes and interfaces to be annotated with special decorators. This means requirement of experimental typescript features for the upstream dependencies.


packages/market-matcher-core/src/CertificateService.ts
```diff
@@ -0,0 +1,111 @@
+import { Demand, PurchasableCertificate } from '@energyweb/market';
+import { Configuration, Unit } from '@energyweb/utils-general';
+import { inject, injectable } from 'tsyringe';
+import * as Winston from 'winston';
+import { TransactionReceipt } from 'web3-core'; // eslint-disable-line import/no-unresolved
```
e00dan: Please remove:
```
// eslint-disable-line import/no-unresolved
```

It should not be used with `web3` imports anymore. If there are errors please install:
```
yarn add web3-core -W
```
to market-matcher-core

e00dan (approved):

PR #317, opened 2019-11-25
fix(utils-general): Alternative ways to extract asset location

e00dan:


kosecki123:
Does the current IREC data scheme was changed or its to support different country? If that’s new country then let’s have a country aware parsing function.

josipbagaric (approved):

PR #408, opened 2019-12-25
feat(origin-ui): input form for registering an organization

e00dan:
localhost_3000_account_organization-register

### DIFF
--- a/packages/utils-general/src/Countries.ts
+++ b/packages/utils-general/src/Countries.ts
@@ -1,0 +1,1496 @@
+export const Countries = [
+    {
+        id: 1,
+        code: 'AF',
+        name: 'Afghanistan',
+        slug: 'afghanistan'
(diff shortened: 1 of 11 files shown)


packages/utils-general/src/Countries.ts
```diff
@@ -0,0 +1,1496 @@
+export const Countries = [
```
josipbagaric, line 1: Do we really want to store this in the repo instead of using a library like https://github.com/michaelwittig/node-i18n-iso-countries?
Is this extracted from the I-REC api?

josipbagaric (approved):

packages/utils-general/src/Countries.ts
e00dan: Yes, it is extracted from I-REC API.

PR #412, opened 2019-12-27
feat(origin-backend): add endpoint to save organizations

e00dan:
- [x] add issuer view for approval


packages/origin-backend/src/api/actions/OrganizationActions.ts
```diff
@@ -0,0 +1,103 @@
...
+    },
+    post: async (req: Request, res: Response) => {
+        console.log(`<POST> Organization`);
+
+        try {
+            const repository = getRepository(Organization);
+
+            const newEntity = new Organization();
+
+            newEntity.address = req.body.address;
```
kosecki123: Probably better approach would be to:

```
constructor(partial: Partial<Organization>) {
    Object.assign(this, partial);
  }
```

and then pass req.body directly.

packages/origin-ui/src/components/Account/OrganizationRegister.tsx
```diff
@@ -150,7 +153,14 @@ export function OrganizationRegister() {
         formikActions.setSubmitting(true);
         dispatch(setLoading(true));
 
-        showNotification('Organization registered.', NotificationType.Success);
+        try {
+            await axios.post(`${environment.BACKEND_URL}/api/Organization`, values);
```
kosecki123: I think we should replace this with a implementation on `origin-backend-client` instead. This is hard to test among other problems.

kosecki123 (approved):

PR #482, opened 2020-01-22
Exchange REST service and order book persistency

kosecki123:
This PR provides the basic implementation of order book exchange for Origin certificates.

Note: This is not complete exchange implementation.


josipbagaric (approved):
👍

packages/exchange-core/src/MatchingEngine.ts
```diff
@@ -1,31 +1,27 @@
...
 
     private asks: List<Ask> = List<Ask>();
 
-    public trades = new Subject<List<Trade>>();
+    public trades = new Subject<List<TradeExecutedEvent>>();
 
     public cancellationQueue = List<string>();
 
-    public submitOrder(order: Order) {
+    constructor(private deviceService: IDeviceService, private locationService: ILocationService) {}
```
e00dan: If these are private and not changed later, then maybe they should be: `private readonly`?

packages/exchange-core/src/test/Matching.test.ts
```diff
@@ -102,21 +105,38 @@ describe('Matching tests', () => {
     const cloneOrder = (order: Order, traded: number) => {
```
e00dan: This should be a generic method, where `T extends Order`. Like:

```
function cloneOrder<T extends Order>(order: T, trader: number)
```

Then you don't need to cast the return of it to either `Bid` or `Ask`, as is currently seen in the code. Some examples from current codebase:

```
const bidsAfter = [cloneOrder(bidsBefore[1], onekWh) as Bid];
// ...
const asksAfter = [cloneOrder(asksBefore[0], onekWh * 3) as Ask, asksBefore[1]];
```

Example implementation of generic method:
```
interface Order {
    order: string;
}

interface Bid extends Order {
    bid: string;
}

function cloneOrder<T extends Order>(order: T, trader: number): T {
    return order;
}

const bid: Bid = null;

const result = cloneOrder(bid, 1); // result will be of type Bid
```

packages/exchange/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+    public location?: string[];
+
+    public deviceVintage?: number;
+}
+```
+
+Where
+
+- `deviceType` - describes the the type of the device for e.g using I-REC types
```
e00dan: Two times `the`

packages/exchange/README.md
```diff
@@ -0,0 +1,76 @@
...
+## Main features
+- Order book matching engine for time, price and product matching
+- ERC 1155 / ERC 1888 compatible
+- Supply / Demand modules 
+
+## Trading product concept
+
+### Definition
+
+`Product` defines the characteristics of the given producing device as well as buyers preferences . Currently it's represented as:
```
e00dan, line 22: Unnecessary space before dot.

packages/exchange/README.md
```diff
@@ -0,0 +1,76 @@
...
+    public location?: string[];
+
+    public deviceVintage?: number;
+}
+```
+
+Where
+
+- `deviceType` - describes the the type of the device for e.g using I-REC types
+- `location` - describes the location of the the device for e.g can be multi-level like Country->Region->Province
```
e00dan: "e. g" should be "e. g." I believe

packages/exchange/src/pods/demand/demand.service.ts
```diff
@@ -0,0 +1,72 @@
...
+                bids: [bid]
+            })
+            .save();
+
+        this.matchingService.submit(bid);
+
+        return this.findOne(userId, demand.id);
+    }
+
+    public async findOne(userId: string, id: string) {
```
e00dan, line 52: Why is `userId` needed to find demand by unique demand UUID?

packages/exchange/src/pods/demand/demand.service.ts
```diff
@@ -0,0 +1,72 @@
...
+        const demand = await this.repository.findOne(id, { where: { userId } });
+        if (!demand) {
+            return null;
+        }
+
+        return { ...demand, trades: demand.trades.map(trade => trade.withMaskedOrder(userId)) };
+    }
+
+    public async getAll(userId: string) {
```
e00dan, line 61: `getAll` should probably be called `getAllForUser` or something like that to better describe it's behavior.

packages/exchange/src/pods/matching-engine/matching-engine.service.ts
```diff
@@ -0,0 +1,91 @@
...
+        const order = this.toOrder(orderEntity);
+        this.matchingEngine.submitOrder(order);
+    }
+
+    public query(product: Product) {
+        // TODO: consider reading this info from DB and filtering
+        return this.matchingEngine.orderBookByProduct(product);
+    }
+
+    @Interval(1000)
```
e00dan, line 56: Cool stuff

e00dan (approved):

packages/exchange-core/src/test/Matching.test.ts
kosecki123: This won't really work in this case because you need to create new instance of the object (using non-default constructor). The way to go is actually to provide clone() function on Bid and Ask class.

packages/exchange/src/pods/demand/demand.service.ts
kosecki123: This provides DB level check if you are an owner of the demand.

packages/exchange-core/src/test/Matching.test.ts
e00dan: I'm not talking about changing any of the method's logic. Only types. I actually provided signature which should be a drop-in replacement:

```
cloneOrder<T extends Order>(order: T, trader: number): T
```

It's worth to try testing it. No need to change any of the method's logic.

PR #505, opened 2020-01-27
Add user to organization

e00dan:
- add linking blockchain account to offchain account functionality via eth_signTypedMessage_v4
- add ability to invite user to organization
- add ability to login/logout
- introduce concept of email/password account that is linked to blockchain account

---

BREAKING CHANGE: require UserClient in configuration
BREAKING CHANGE: change CREATE_ACCOUNT, remove ADD_CONSUMING_DEVICE

- utils-general: add userClient requirement to configuration offchain data properties
- utils-general: add typed data sign and recover methods, sign method works with MetaMask and passing private key to it

- origin-backend-core: add OrganizationInvitation structure, re-orgnize User and Organization types to better reflect data returned and passed to backend

- origin-backend-client: remove mocks from package - and create new package for them - origin-backend-client-mocks; this is due to hitting circular reference error (by adding reference to utils-general, for signing, in UserClientMock)
- origin-backend-client: introduce RequestClient, to provide abstraction over direct calls to `axios`

- origin-backend: better use types like`IUserWithRelationsIds` to correctly describe data underneath; for example if you load entity with `loadRelationsIds` option, typeorm incorrectly returns it's type - it says that relation is full object, while it's only number representing id
- origin-backend: add `@UserDecorator()` to automatically fill logged user account in endpoints
- origin-backend: add `blockchainSignedMessage` and `blockchainAccountAddress` properties

- migrations: change CREATE_ACCOUNT, add CREATE_ORGANIZATION, remove ADD_CONSUMING_DEVICE
- migrations: remove roles that weren't used from demo-config.json
- migrations: improve error reason logging

- user-registry: remove all precise-proofs user offchain properties
- user-registry: sync new offchain account and attach it as `user.information: IUser` property


packages/event-listener/src/test/deployDemo.ts
```diff
@@ -211,98 +206,92 @@ export class Demo {
...
+            notifications: false
         };
 
         await MarketUser.createMarketUser(
             marketLogicPropsOnChain,
             marketLogicPropsOffChain,
-            this.conf
+            this.conf,
+            null,
+            null,
```
josipbagaric: Is there no way to avoid forwarding `null`?

packages/market/src/blockchain-facade/MarketUser.ts
```diff
@@ -37,12 +39,25 @@ export class Entity extends User.Entity {
 
         return new Entity(this.id, this.configuration).sync();
     }
+
+    async sync(): Promise<Entity> {
+        return super.sync();
```
josipbagaric, line 44: Is this really needed?

packages/user-registry/src/blockchain-facade/Users/User.ts
```diff
@@ -1,37 +1,28 @@
...
-    firstName?: string;
-    surname?: string;
-    street?: string;
-    number?: string;
-    zip?: string;
-    city?: string;
-    country?: string;
-    state?: string;
-}
+export type IUserOffChainProperties = {};
```
josipbagaric, line 7: Can this be removed then?

josipbagaric (changes requested):

packages/event-listener/src/test/deployDemo.ts
josipbagaric: Applies to multiple places where this is done.

packages/market/src/blockchain-facade/MarketUser.ts
e00dan: Previously `MarketUser.sync()` returned `User.Entity` instead of `MarketUser.Entity`. This is a fix that improves types and correctly returns `MarketUser.Entity` from `MarketUser.Entity.sync()` function.

packages/user-registry/src/blockchain-facade/Users/User.ts
e00dan: I have experienced problems with trying to remove user offchain properties and `MarketUser` actually uses offchain properties (thus, for example, a need to leave logic in Solidity UserLogic).

I'd rather have this dummy type, as user-registry is going to change a lot soon and I see no harm in leaving 1 interface that is empty object - especially that is still referenced in multiple places where `User.Entity` is created - and that is problematic to due to being extended from `MarketUser.Entity`.

One of the problems I've experienced was with `prepareEntityCreation` where user off chain properties were empty.

packages/event-listener/src/test/deployDemo.ts
e00dan: I'll try to do that and see if it improves codebase

josipbagaric (approved):

packages/market/src/blockchain-facade/MarketUser.ts
```diff
@@ -58,36 +72,63 @@ export const createMarketUser = async (
...
 
     if (!successCreateUser || !successSetRoles) {
         await user.deleteFromOffChainStorage();
         throw new Error('createMarketUser: Saving on-chain data failed. Reverting...');
     }
 
+    if (registerData) {
+        const userClient = configuration?.offChainDataSource?.userClient;
+
+        const userOffchain = await userClient.register({
```
kosecki123: I guess we should remove test default data?

packages/origin-backend/src/pods/organization/organization.controller.ts
```diff
@@ -16,32 +20,71 @@ import {
...
     @Get('/:id')
     async get(@Param('id') id: string) {
-        console.log(`<GET> Organization/${id}`);
-
-        const existingEntity = await this.organizationRepository.findOne(id);
+        const existingEntity = await this.organizationRepository.findOne(id, {
+            loadRelationIds: true
+        });
 
         if (!existingEntity) {
```
kosecki123, line 89: You can replace that with interceptor like:

```
import {
    Injectable,
    NestInterceptor,
    ExecutionContext,
    CallHandler,
    NotFoundException
} from '@nestjs/common';
import { Observable } from 'rxjs';
import { map } from 'rxjs/operators';

@Injectable()
export class EmptyResultInterceptor implements NestInterceptor {
    intercept(context: ExecutionContext, next: CallHandler): Observable<any> {
        return next.handle().pipe(
            map(data => {
                if (!data) {
                    throw new NotFoundException();
                }
                return data;
            })
        );
    }
}
```

kosecki123 (changes requested):

packages/origin-backend/src/pods/organization/organization.controller.ts
e00dan: This would affect a major part of codebase, I'd prefer not to do it as part of this pull request which already contains big changes.

kosecki123 (approved):

packages/device-registry/src/test/DeviceFacade.test.ts
```diff
@@ -12,331 +12,329 @@ import {
...
+
+    //             assert.deepEqual(reads, [
+    //                 {
+    //                     energy: 100,
+    //                     timestamp: SM_READ_TIMESTAMP
+    //                 }
+    //             ]);
+    //         });
+    //     });
+    // });
```
josipbagaric, line 339: Not sure how I missed this, but these tests shouldn't be commented out.

PR #507, opened 2020-01-28
Device and Demand clients

josipbagaric:
- Added a `/Device` endpoint to the backend
- Added a DeviceClient to communicate with the backend
- Removed the notion of a ConsumingDevice and DeviceTypes - all devices are now producing devices
- Removed storing the precise proof on-chain (part of the transition to the off-chain device registry)
- Use `OffChainDataSource` class instead of individual clients for storing different entities on the backend
- Refactored `Demand` to fit structurally into the backend and removed all on-chain references
- Removed all unused code for `Agreements` and `Supply`


packages/event-listener/package.json
```diff
@@ -53,7 +53,8 @@
         "winston": "3.2.1"
     },
     "devDependencies": {
-        "@energyweb/origin-backend-client-mocks": "0.0.1"
+        "@energyweb/origin-backend-client-mocks": "0.0.1",
+        "@energyweb/origin-backend-core": "^0.1.1"
```
e00dan: Please use fixed version: "0.1.1"

packages/market-matcher-core/package.json
```diff
@@ -42,6 +42,7 @@
         "winston": "3.2.1"
     },
     "devDependencies": {
+        "@energyweb/origin-backend-core": "^0.1.1",
```
e00dan: Please use fixed version

packages/market-matcher/package.json
```diff
@@ -54,6 +54,7 @@
         "winston": "3.2.1"
     },
     "devDependencies": {
+        "@energyweb/origin-backend-core": "^0.1.1",
```
e00dan: Please use fixed version

packages/origin-backend-client/src/DeviceClient.ts
```diff
@@ -0,0 +1,40 @@
...
+        this.baseURL = _baseUrl;
+    }
+
+    private get endpoint() {
+        return `${this.baseURL}/Device`;
+    }
+
+    public async getById(id: number): Promise<IDevice> {
+        const url = `${this.endpoint}/${id}`;
+        const { data } = await axios.get(url);
```
e00dan: Please use `RequestClient` class from backend, instead of invoking axios directly. One of the reasons for using `RequestClient` is better way to configure request parameteres, like authentication token - for authorized XHR calls.

e00dan (approved):

PR #509, opened 2020-01-30
@energyweb/exchange integration with @energyweb/origin-backend

kosecki123:
This PR provides the integration of the @energyweb/exchange endpoints with origin-backend using common authentication system.


josipbagaric (approved):

packages/exchange/src/app.module.ts
```diff
@@ -61,6 +62,15 @@ const getEnvFilePath = () => {
...
     ],
+    // exports: [
+    //     TypeOrmModule,
+    //     ScheduleModule,
+    //     MatchingEngineModule,
+    //     TradeModule,
+    //     OrderModule,
+    //     DemandModule,
+    //     OrderBookModule
+    // ],
```
josipbagaric: Should this be commented out?

josipbagaric (approved):

e00dan (approved):

packages/exchange/src/app.module.ts
kosecki123: Yeah. The should be removed, as we are not importing them one-by-one by as a whole module.

PR #536, opened 2020-02-03
WebSocket Event Server

josipbagaric:
### Changes
- A new WebSocket-based off-chain eventing system
- Moved all the data clients to `OffChainDataSource`
- Partially move `Device`s off-chain
- Make `Demand` entirely off-chain
- Remove `Agreement` and `Supply`
- Listen to off-chain events in `event-listener`, `market-matcher` and the UI


packages/device-registry/src/test/DeviceFacade.test.ts
```diff
@@ -113,17 +104,14 @@ describe('Device Facade', () => {
...
+                lastSmartMeterReadFileHash: 'lastSmartMeterReadFileHash'
             };
 
-            const devicePropsOffChain: ProducingDevice.IOffChainProperties = {
+            const devicePropsOffChain: IDevice = {
+                status: DeviceStatus.Active,
                 operationalSince: 0,
                 capacityInW: 10,
-                country: 'Thailand',
+                country: 221,
```
kosecki123: What does 221 means?

josipbagaric: That's the ID of the country in the Countries array. This was just for tests, I will replace it with `Countries.find(c => c.name === 'Thailand').id`

packages/origin-ui-core/src/components/AddDevice.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { IRECDeviceService } from '@energyweb/utils-general';
+import { IRECDeviceService, Countries } from '@energyweb/utils-general';
```
kosecki123, line 2: We also have a Country Repository https://github.com/energywebfoundation/origin/blob/b6d6609f0031c51e7a6943590b60607e1035ede4/packages/origin-backend/src/pods/country/country.controller.ts

Which one we should use? The static one from utils-general or the one from DB?
@Kuzirashi

kosecki123 (approved):

e00dan (approved):

PR #556, opened 2020-02-08
fix(issuer): double requesting for the same generation period

josipbagaric:
Fixes an issue where a user was able to request issuance of a certificate for which it has already been issued.

### DIFF
--- a/packages/issuer/contracts/PublicIssuer.sol
+++ b/packages/issuer/contracts/PublicIssuer.sol
@@ -24,6 +24,8 @@ contract PublicIssuer is Initializable, Ownable {
     mapping(uint256 => RequestIssue) public requestIssueStorage;
     mapping(uint256 => uint256) public certificateToRequestStorage;
 
+    mapping(string => uint256) private deviceLatestGenerationTimestamp;
+
     function initialize(address _registry) public initializer {
         require(_registry != address(0), "initialize: Cannot use address 0x0 as registry address.");
         registry = Registry(_registry);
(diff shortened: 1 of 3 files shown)


kosecki123 (changes requested):
I'm not sure if should disable requesting the certs for the same periods. Possibly you can request different kind of certificate from the same device for the same period of time. 

Let's discuss and keep this open until that.

kosecki123:
> I'm not sure if should disable requesting the certs for the same periods. Possibly you can request different kind of certificate from the same device for the same period of time.
> 
> Let's discuss and keep this open until that.

2nd thought is, since single issuer contract is bound to a single certificate topic (https://github.com/energywebfoundation/origin/blob/daf4966edd090940724acb53359b98be33d1f140/packages/issuer/contracts/PublicIssuer.sol#L13) it's actually a good way to prevent spamming with same period requests.

JosephBagaric:
> > I'm not sure if should disable requesting the certs for the same periods. Possibly you can request different kind of certificate from the same device for the same period of time.
> > Let's discuss and keep this open until that.
> 
> 2nd thought is, since single issuer contract is bound to a single certificate topic (
> 
> https://github.com/energywebfoundation/origin/blob/daf4966edd090940724acb53359b98be33d1f140/packages/issuer/contracts/PublicIssuer.sol#L13
> 
> ) it's actually a good way to prevent spamming with same period requests.

@kosecki123 this PR only stops

PR #563, opened 2020-02-10
relative path to root .env file fixed

ChangoBuitrago:


e00dan:
Hello, sorry for getting back to you that late. 

We do currently use solar-simulator with existing config and it seems to work properly. We're also using this file for 3 or 4 pilots. That's why I'd like to ask, if you could explain more on the context of this change, and, ideally if you could provide some tests if you think something's not working.

On my side I can provide some scenarios that should be working (which seem to work now):

- ENV file should be correctly read by running solar-simulator server and consumer process
- ENV file should be correctly run by mockreadings command (both top-level file, and also sub-level mockReadingsWorker)
- ENV file should be correctly read when running `yarn start` in solar-simulator
- ENV file should be correctly read when installing @energyweb/solar-simulator as separate package and running its executable (example command that runs it like that is: `npx @energyweb/solar-simulator -e ../../../../../../.env -c ../../../../../../config/solar-simulator/config.json -d ../../../../../../config/solar-simulator/data.csv`)

I know it's a lot of cases, but that's why I'm also hesitant on accepting the change without (ideally automated) tests.

Thank you for your contribution!

ChangoBuitrago:
Hi @Kuzirashi. Thanks for your message. I'm also sorry for my late reply. Was sick in bed all week.

Most of your scenarios seems to run directly from the command line. If that's the case, they will work fine because they will all go through the following lines of code:
```
if (require.main === module) {
    dotenv.config({
        path: '../../.env'
    });
   ...
}
```
In my case, I'm just running `npm run 'run:origin'` script from the app root dir. Which will include the solar-simulator as a module and avoid the above thread. 
To replicate the issue:
1. Set the `SOLAR_SIMULATOR_DEPLOY_PAST_READINGS` variable in your root `.env` file to `true`.
2. From the app root dir run: `npm run 'run:origin'`
3. If the issue appears, past readings will **not be loaded**. 

Regards,

PR #566, opened 2020-02-11
fix(issuer): Check generation time periods before requesting issuance

josipbagaric:


packages/issuer/contracts/AbstractIssuer.sol
```diff
@@ -55,6 +57,14 @@ contract AbstractIssuer is Initializable, Ownable {
     }
 
     function requestIssueFor(bytes memory _data, address _owner) public returns (uint) {
+        (uint256 from, uint256 to, string memory deviceId) = decodeIssue(_data);
+        require(to > from, "Generation period invalid. 'to' is lower than 'from'");
+        require(to <= block.timestamp, "Generation period invalid. 'to' is higher than now");
+        require(
+            from >= deviceLatestGenerationTimestamp[deviceId],
```
e00dan: I don't think we want to have it sequential, user should be able to request certificates for arbitrary time range.

e00dan (approved):

PR #567, opened 2020-02-11
feat: change certification requests to be off-chain evidence based

e00dan:
ezgif com-video-to-gif (11)

BREAKING CHANGE: move certificate requests logic off-chain
BREAKING CHANGE: remove getCertificationRequestEvents() from PurchasableCertificate


packages/origin-backend-client-mocks/src/OffChainDataSourceMock.ts
```diff
@@ -8,36 +8,44 @@ import {
...
+
     configurationClient: IConfigurationClient = new ConfigurationClientMock();
+
     userClient: IUserClient = new UserClientMock();
+
     eventClient: IEventClient = new EventClientMock();
-    
+
     demandClient: IDemandClient;
+
```
josipbagaric, line 35: Am I the only one who finds this rule annoying? 😅

josipbagaric (approved):

packages/origin-backend-client-mocks/src/OffChainDataSourceMock.ts
e00dan: I'm neutral towards it. ; p I can format it with one keyword so doesn't make much difference for me. ;p

PR #573, opened 2020-02-12
Feat/exchange deposit watcher

kosecki123:
This PR features a deposit watcher for ERC1888 tokens.

**This does not contain proper way of reading the device id from the token, since the final structure of `bytes` field is not yet know. This was annotated with //TODO comment.**


packages/exchange/src/pods/account/account.service.ts
```diff
@@ -11,6 +11,8 @@ import { Account as AccountEntity } from './account.entity';
 
 @Injectable()
 export class AccountService {
+    private readonly logger = new Logger(AccountService.name);
```
e00dan, line 14: 👏

josipbagaric (approved):

packages/exchange/src/pods/asset/asset.service.ts
```diff
@@ -17,11 +19,12 @@ export class AssetService {
     }
 
     public async createIfNotExist(asset: AssetDTO, transaction?: EntityManager) {
+        this.logger.debug(`Requested asset ${JSON.stringify(asset)}`);
         if (transaction) {
             return this.create(asset, transaction);
         }
 
-        return this.connection.transaction(tr => this.create(asset, tr));
+        return this.connection.transaction(async tr => this.create(asset, tr));
```
e00dan, line 27: Unnecessary async, as this.create already returns a promise

e00dan (approved):

PR #595, opened 2020-02-21
fix(issuer): Commitment updating, Private Transfers and Claims

josipbagaric:
- Fix an issue where updating claims would fail due to using extended merkle trees on one side of the check
- Fix checks for private transfers
- Fix proof checking ownership for private claiming (depends on https://github.com/energywebfoundation/precise-proofs/pull/7)
- Allow approving private transfers and public migrations only by the issuer
- Different `isPrivate` check for certificates
- Use the main `rootHash` in precise proofs instead of the extended merkle tree root (to remain compatible with on-chain precise proof checks)

### DIFF
--- a/packages/issuer/contracts/Issuer.sol
+++ b/packages/issuer/contracts/Issuer.sol
@@ -156,7 +156,7 @@ contract Issuer is Initializable, Ownable {
         uint _requestId,
         bytes32 _commitment,
         bytes memory _validityData
-    ) public returns (uint256) {
+    ) public onlyOwner returns (uint256) {
         IssuanceRequest memory request = issuanceRequests[_requestId];
         require(request.isPrivate, "approveIssuancePrivate: can't approve public certificates using commitments");
 
@@ -172,7 +172,7 @@ contract Issuer is Initializable, Ownable {
         return certificateId;
     }
 
-    function issuePrivate(address _to, bytes32 _commitment, bytes memory _data) public returns (uint256) {
+    function issuePrivate(address _to, bytes32 _commitment, bytes memory _data) public onlyOwner returns (uint256) {
         uint256 requestId = requestIssuanceFor(_data, _to, true);
 
         return approveIssuancePrivate(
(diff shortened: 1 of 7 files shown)


packages/issuer/contracts/Issuer.sol
```diff
@@ -213,6 +218,10 @@ contract Issuer is Initializable, Ownable {
 	*/
 
 	function requestMigrateToPublic(uint _certificateId, bytes32 _ownerAddressLeafHash) external returns (uint256) {
+        bool existsAlready = _migrationRequestExists(_certificateId);
```
kosecki123, line 221: `bool alreadyExists` or even `bool exists` ?

packages/issuer/contracts/Issuer.sol
```diff
@@ -227,26 +236,40 @@ contract Issuer is Initializable, Ownable {
         return id;
 	}
 
-	// TO-DO: only Issuer
+    function getMigrationRequestId(uint _certificateId) external onlyOwner returns (uint256) {
```
kosecki123, line 239: I think that each certificate can have multiple requests to go to public. So for e.g 1 private owner exposes it now while other performs this at later time.

josipbagaric: True, this should be changed. Thanks.

kosecki123: In this case if we want to disallow same migration to be requested. We can store hash of the request (abi.encode(id, owneradd)) or possibly just the  _ownerAddressLeafHash  mapped to certificate id.

mapping(certId => _ownerAddressLeafHash)

kosecki123 (changes requested):
Requesting changes based on the comments.

packages/issuer/contracts/Issuer.sol
kosecki123: Also, since this method was lineary growing in the execution time, it would be more and more expensive for users to call. We should if possible to optimize such a O(N) iterations with O(1) mapping calls.

kosecki123 (approved):

josipbagaric:
Implemented in https://github.com/energywebfoundation/origin/pull/663. Closing.

PR #603, opened 2020-02-25
Feat: Smart meter reading adapters

josipbagaric:
- Default smart meter readings to be stored off-chain instead of off-chain
- Allow adapters for getting/setting smart meter readings from different sources


packages/origin-backend/src/app.module.ts
```diff
@@ -34,53 +34,59 @@ import { OrganizationModule } from './pods/organization/organization.module';
...
-        CertificateModule
-        // TODO: enable exchange endpoints
-        // ExchangeAppModule
-    ],
-    controllers: [AppController]
-})
-export class AppModule {}
+@Module({})
+export class AppModule {
+    static register(smartMeterReadingsAdapter: ISmartMeterReadingsAdapter): DynamicModule {
```
e00dan, line 43: Could you elaborate on this change? I don't understand it

e00dan (approved):

kosecki123:
I suggest we move the reads related code into separate adapter instead of removing it. Smart contract functionally can also stay as it is, although further refactoring would be great.

josipbagaric:
> I suggest we move the reads related code into separate adapter instead of removing it. Smart contract functionally can also stay as it is, although further refactoring would be great.

Tried this but faced some issues with Nest when initializing the "default" adapter. Because it needs to be initialized before starting the main AppModule, there is no way for it to interact with the Device repository (where default readings would be stored).

What I've done instead in this PR is that origin-backend by default stores smart meter readings directly in Device entity, and this can be overridden by an adapter.

kosecki123 (approved):

PR #606, opened 2020-02-27
feat(origin-ui-core): Bids and Asks UI list component

e00dan:
- Configure Storybook
- Move OrderBookOrder DTO from exchange to exchange-core
- Add information to exchange readme on exact commands used to run PostgreSQL and pgAdmin


packages/exchange-core/src/index.ts
```diff
@@ -5,3 +5,4 @@ export { Ask } from './Ask';
 export { Product } from './Product';
 export { DeviceVintage } from './DeviceVintage';
 export { Operator } from './Operator';
+export * from './order-book-order.dto';
```
kosecki123, line 8: Let not move those DTO objects to core project. They are specific for exchange and clients and meant for the data serialization/deserialization.

I think it's fine to have either a copy of DTO object in the UI or create exchange-client package and put it there.

kosecki123 (changes requested):

PR #608, opened 2020-02-28
feat(origin-ui-core): Bids and Asks UI list component

e00dan:
- Configure Storybook
- Move OrderBookOrder DTO from exchange to exchange-core
- Add information to exchange readme on exact commands used to run PostgreSQL and pgAdmin


josipbagaric (approved):

packages/origin-ui-core/src/components/exchange/order-book-order.dto.ts
```diff
@@ -0,0 +1,19 @@
...
+export class OrderBookOrderDTO {
+    price: number;
+
+    volume: string;
+
+    product: Product;
+
+    userId: string;
+
+    public static fromOrder(order: Bid | Ask, userId?: string): OrderBookOrderDTO {
```
kosecki123: You don't need that method. Remove it, this will also allow you to remove @energyweb/exchange-core dependency.

kosecki123 (changes requested):
Rm @energyweb/exchange-core dependency

kosecki123 (approved):

PR #664, opened 2020-03-11
feat: implement additional ids for registry

e00dan:


josipbagaric (approved):

kosecki123 (changes requested):
In general adding new IDs as a datafields is not really an optimal way. As we hardcoding some specific concepts into the device definitions.

Let's think about the solution where we have more generic approach.

kosecki123:
Ideally this would be implemented as:

1) There is configuration item (in `.env` or DB) that defines possible external id types like `['LIVE_PRODUCTION', 'ISSUER_ID', 'SOMECOOL_ID'] and this is available system wide - important thay we keep this a "dynamic" data, not a enum.
2) Device entity stores a json field: externalId: [{id: IDVALUE, type: TYPE}] for e.g `[{id: '1234', type: 'LIVE_PRODUCTION'}]`
3) API endpoint /device-reads/{originId}/(optional)TYPE
4) UI registration form creates fields dynamically based on the 1) and stores the output as a JSON

In summary:
1) Origin SDK - works agains some 'strings'
2) Project A - which needs 3 different TYPES of ID, would configure it as 1) then store them in migration or register via UI, reads adapter will receive whole Device object including this ID and query the destination using correct ID

josipbagaric:
> Ideally this would be implemented as:
> 
> 1. There is configuration item (in `.env` or DB) that defines possible external id types like `['LIVE_PRODUCTION', 'ISSUER_ID', 'SOMECOOL_ID'] and this is available system wide - important thay we keep this a "dynamic" data, not a enum.
> 2. Device entity stores a json field: externalId: [{id: IDVALUE, type: TYPE}] for e.g `[{id: '1234', type: 'LIVE_PRODUCTION'}]`
> 3. API endpoint /device-reads/{originId}/(optional)TYPE
> 4. UI registration form creates fields dynamically based on the 1) and stores the output as a JSON
> 
> In summary:
> 
> 1. Origin SDK - works agains some 'strings'
> 2. Project A - which needs 3 different TYPES of ID, would configure it as 1) then store them in migration or register via UI, reads adapter will receive whole Device object including this ID and query the destination using correct ID

Good approach. Will continue with implementing this, defining the IDs in `demo-config.json` instead of `.env`.

kosecki123:
> > Ideally this would be implemented as:
> > 
> > 1. There is configuration item (in `.env` or DB) that defines possible external id types like `['LIVE_PRODUCTION', 'ISSUER_ID', 'SOMECOOL_ID'] and this is available system wide - important thay we keep this a "dynamic" data, not a enum.
> > 2. Device entity stores a json field: externalId: [{id: IDVALUE, type: TYPE}] for e.g `[{id: '1234', type: 'LIVE_PRODUCTION'}]`
> > 3. API endpoint /device-reads/{originId}/(optional)TYPE
> > 4. UI registration form creates fields dynamically based on the 1) and stores the output as a JSON
> > 
> > In summary:
> > 
> > 1. Origin SDK - works agains some 'strings'
> > 2. Project A - which needs 3 different TYPES of ID, would configure it as 1) then store them in migration or register via UI, reads adapter will receive whole Device object including this ID and query the destination using correct ID
> 
> Good approach. Will continue with implementing this, defining the IDs in `demo-config.json` instead of `.env`.

Good point! Agree.

PR #665, opened 2020-03-11
fix(deps): add commander@4.1.1 to migration

chonla:
This PR adds missing dependency, commander@4.1.1, from migration package.


kosecki123 (approved):
LGTM

kosecki123:
FYI: We typically lock the versions, so instead of `~4.1.1` we use `4.1.1`. All asterisks will be locked automatically by Renovate tool later anyways.

PR #699, opened 2020-03-17
fix: make ContractsLookup part of Configuration

e00dan:


josipbagaric (approved):

packages/exchange/src/pods/runner/runner.service.ts
```diff
@@ -28,7 +29,18 @@ export class RunnerService implements OnModuleInit {
     }
 
     public async fetchDeviceTypesAndInit() {
-        const getDeviceTypes = () => this.deviceTypesService.get();
+        const getDeviceTypes = async () => {
+            let deviceTypes: IDeviceType[] = [];
+
+            try {
+                const configuration = await this.originBackendConfigurationService.get();
```
kosecki123: Previously, device-type were passed as a type, here we passing the config client to fetch. So this effectively means the backend is asking over https itself to get the config?

PR #700, opened 2020-03-17
fix: remove MarketUser from UI

e00dan:
- Remove `@energyweb/market` usage from UI
- Add creating `devices` to `@energyweb/migrations`
- Remove in-browser private key account selector
- Rewrite `CertificateTable` to functional component


packages/origin-backend-core/src/User.ts
```diff
@@ -1,5 +1,13 @@
 import { IOrganization } from './Organization';
 
+type Currency = string;
+
+export interface IAutoPublishConfig {
+    enabled: boolean;
+    currency: Currency;
+    priceInCents: number;
```
josipbagaric, line 8: Is this still needed? Will certificates still be autopublished in the new exchange?

josipbagaric (approved):

packages/origin-backend-core/src/User.ts
kosecki123: Yes, we will retain such a functionality.

kosecki123 (approved):
LGTM

PR #708, opened 2020-03-19
[Feat] Multi-owner private certificates

josipbagaric:
Screenshot 2020-03-24 at 12 01 13

### Features
- [x] Multi-owner private certificates
- [x] Support storing OwnershipCommitments off-chain
- [x] Allow the Issuer to approve private transfers

### Fixes
- [x] Re-enable smart contract commitment checks
- [x] On-chain `isCertificatePrivate` check

### Chore
- [x] Remove `BlockchainDataModelEntity` and do a tighter integration with Certificates and CertificationRequests

## Limitations
- Only one private transfer request per certificate can exist at any time


e00dan (approved):

packages/origin-backend/src/pods/certificate/certificate.controller.ts
```diff
@@ -37,4 +47,92 @@ export class CertificateController {
     ): Promise<CertificationRequestOffChainData> {
         return this.certificationRequestRepository.findOne(id);
     }
+
+    @Get('/:id')
```
kosecki123, line 51: Should use auth system to authenticate user and allow only users commitments to be served.

kosecki123 (changes requested):
LGTM but there are not a single test for this code. I think this should be tested thoroughly.

josipbagaric:
> LGTM but there are not a single test for this code. I think this should be tested thoroughly.

Sure, there are old tests which pass - but will add new tests to cover more cases.

kosecki123 (approved):

PR #722, opened 2020-03-24
chore: replace email notifications services architecture

e00dan:
- Create EventsService that dispatches event both to NotificationsService (sending e-mails) as well as to WebSockets
- Add following e-mail notifications: `ORGANIZATION_INVITATION, ORGANIZATION_STATUS_CHANGED`

This PR lays the foundation for implementing rest of the notifications.


packages/origin-backend/src/pods/mail/mail.module.ts
```diff
@@ -0,0 +1,44 @@
...
+            useFactory: async (configService: ConfigService) => {
+                return {
+                    transport: mandrillTransport({
+                        auth: {
+                            apiKey: configService.get<string>('MANDRILL_API_KEY')
+                        },
+                        // eslint-disable-next-line @typescript-eslint/camelcase
+                        from_email: configService.get<string>('EMAIL_FROM'),
+                        // eslint-disable-next-line @typescript-eslint/camelcase
+                        from_name: 'Energy Web Origin',
```
kosecki123: I think all of those string should be configurable.

packages/origin-backend/src/pods/mail/mail.service.ts
```diff
@@ -0,0 +1,55 @@
...
+        try {
+            const result: IMandrillMailSendStatus = await this.mailerService.sendMail(
+                sendMailOptions
+            );
+
+            if (result.messageId) {
+                const allSucceeded = result.accepted.reduce(
+                    (a, b) => a && b.status === 'sent',
+                    true
+                );
```
kosecki123: Use Array.prototype.every() instead.

packages/origin-backend/src/pods/mail/mail.service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    constructor(private readonly mailerService: MailerService) {}
+
+    async send(sendMailOptions: ISendMailOptions) {
+        try {
+            const result: IMandrillMailSendStatus = await this.mailerService.sendMail(
+                sendMailOptions
+            );
+
+            if (result.messageId) {
```
kosecki123: What does this case means? We silently skipping it. Maybe should be logged?

kosecki123: Also, possibly better as 

```
if (!result.messageId) {
 return true
}
```

packages/origin-backend/src/pods/notification/notification.service.ts
```diff
@@ -0,0 +1,92 @@
...
+    };
+
+    constructor(private readonly mailService: MailService) {}
+
+    async handleEvent(event: NewEvent) {
+        try {
+            assertIsSupportedEvent(event);
+
+            return this.handlers[event.type](
+                (event.data as TSupportedNotificationEvent['data']) as any
```
kosecki123, line 64: Should not cast to `any`?

kosecki123 (approved):
LGTM, some minor code improvements would be nice.

packages/origin-backend/src/pods/mail/mail.module.ts
```diff
@@ -0,0 +1,44 @@
...
+                    defaults: {
+                        from: `"Energy Web Origin" <${configService.get<string>('EMAIL_FROM')}>`
+                    },
+                    template: {
+                        dir: `${__dirname}/templates`,
+                        adapter: new HandlebarsAdapter(),
+                        options: {
+                            strict: true
+                        }
+                    }
```
josipbagaric, line 37: Are these templates used in any way?

josipbagaric (approved):

packages/origin-backend/src/pods/mail/mail.service.ts
e00dan: This case means if `result.messageId` then we're probably using Mandrill transport. Email transporters in nodemailer seem to have different responses unfortunately, and this needs to be abstracted away,

packages/origin-backend/src/pods/notification/notification.service.ts
e00dan: In this case, because these values are dynamic, I haven't found a good way how to do it yet. The surrounding code guards things nicely though.

packages/origin-backend/src/pods/mail/mail.module.ts
e00dan: They aren't. but can be, if we decide to use them.

PR #723, opened 2020-03-25
Generation time matching

kosecki123:
This PR provides:

- [x] matching engine respecting generation time
- [x] order book filters
- [x] exchange API filters integration
- [x] generation time reading from token meta-data
- [x] asset entity storing generation time
- [x] updated example.sql file

Note: Since this PR Ask object requires generation time data to be available in the Product object. Previously created Asks will not be correctly injected into matching engine causing exchange start to fail. Please truncate all existing orders and assets or edit manually.


josipbagaric (approved):
LGTM

PR #731, opened 2020-03-26
Feat: Origin-backend on Postgres

josipbagaric:
- Transition from SQLite to Postgres
- Remove old `Demand` structure
- Add `createdAt` and `updatedAt` to all entities
- Add an initial migration example
- Automatically drop schemas when not running in production


kosecki123 (changes requested):
All good but you need to update `.travis.yml`

https://github.com/energywebfoundation/origin/blob/2d474ab448b632944f171cd510e3fa5c9c34acd7/.travis.yml#L9

kosecki123 (approved):

e00dan (approved):

PR #760, opened 2020-03-30
Migrations using SQL files

josipbagaric:
## New migration workflow
0. You need to have Postgres running with a database called `origin`.
1. `yarn run:ganache` (or use Volta)
2. `yarn migrate:demo`. This will redeploy the contracts every time. Use `migrate:prod` in production.
3. `yarn run:backend`
4. `yarn run:ui`

**OR**

5. `yarn run:origin` to run all of the above


kosecki123 (approved):
Great work!

e00dan (approved):

PR #763, opened 2020-04-01
Exchange based demands

kosecki123:
This PR provides the API for demands management, API for orders cancellation

Features:
- [x] Demand creation
- [x] Order cancellation POST /:id/cancel
- [x] Order GET /:id
- [x] Cancellation/edits, demand states

Known issue where consecutive pause/resume/pause with less that ~1sec demand might end up with orders having wrong states #765


packages/exchange/test/demand.e2e-spec.ts
```diff
@@ -118,4 +145,115 @@ describe('Demand orders trading', () => {
...
+
+                expect(bid1.price).toBe(createDemandWith2Bids.price);
+                expect(bid2.price).toBe(createDemandWith2Bids.price);
+
+                expect(bid1.demandId).toBe(created.id);
+                expect(bid2.demandId).toBe(created.id);
+            });
+    });
+
+    it('should be able to cancel demand', async () => {
```
josipbagaric, line 170: Cancelling === Pausing demand?

packages/utils-general/src/blockchain-facade/EnumExports.ts
```diff
@@ -15,3 +15,9 @@ export enum Unit {
     GW = 1e9,
     GWh = 1e9
 }
+
+export enum DemandStatus {
```
josipbagaric, line 19: Can this be moved to the exchange-core package?

josipbagaric (approved):
LGTM. Only have a few questions, but can otherwise be merged.

packages/exchange/test/demand.e2e-spec.ts
kosecki123: Yes, typo, should be "pause demand"

packages/utils-general/src/blockchain-facade/EnumExports.ts
kosecki123: I though so too, but exchange-core does not have any notion of the demand, it's bids, asks, orders, trades. Demand is a collection of bids.

PR #764, opened 2020-04-01
feat: add posting certificates for sale

e00dan:


packages/issuer/src/blockchain-facade/CertificationRequest.ts
```diff
@@ -41,11 +40,15 @@ export class Entity extends PreciseProofEntity implements ICertificationRequest
...
+        const device = await this.configuration.offChainDataSource.deviceClient.getByExternalId({
+            type: process.env.ISSUER_ID || 'Issuer ID',
+            id: decodedData['2']
+        });
 
         this.owner = issueRequest.owner;
         this.fromTime = Number(decodedData['0']);
         this.toTime = Number(decodedData['1']);
-        this.device = Number(decodedData['2']);
+        this.device = device?.id || decodedData['2'];
```
kosecki123: I think this is unnecessary and will confuse. We should always assume the correct ID is stored in the token.data field.

kosecki123 (approved):
Besides the commented one, LGTM

PR #771, opened 2020-04-02
fix(exchange-core): dont crash on filtering unspecified bids

e00dan:


kosecki123:
This is good fix, but same if would have to be applied to other filters (vintage, location...).

kosecki123 (approved):
Good

PR #772, opened 2020-04-02
chore(migrations): include create_schema.sql in the NPM release

josipbagaric:


kosecki123:
This seems weird, why do you need to config defaults?

josipbagaric:
> This seems weird, why do you need to config defaults?

Well, we don't need all of the files - but `create_tables.sql` would have to be copied to every other implementation of Origin, as it stays the same regardless.

I could only move `create_tables.sql`, but decided to move them all to have a reference file for each of the required files.

e00dan (approved):

PR #778, opened 2020-04-03
fix(origin-ui-core): display notification when users tries to register device without organization

e00dan:


josipbagaric (approved):

packages/localization/src/languages/en.json
```diff
@@ -153,7 +153,9 @@
             "feedback": {
                 "noChangesMade": "No changes have been made.",
                 "unknownError": "Unknown error",
-                "createdByNoun": "Created by {fullName} from the Noun project"
+                "createdByNoun": "Created by {fullName} from the Noun project",
```
kosecki123, line 156: What is that ? Noun project?

e00dan: They created free SVG icon that we use, for example for device type display in device details, but we have to reference them.

PR #807, opened 2020-04-07
Grid operators

kosecki123:
This PR provides:

- [x] Grid operators as a `string[]` in the configuration
- [x] Grid operator field in `device` entity
- [x] Matching engine matching 
- [x] Bids product extension


e00dan (approved):
Looks good, please smoke test exchange UI though to see if it works with new filter for grid operators before merging

PR #820, opened 2020-04-10
Remove websockets gateway from origin backend

kosecki123:
This PR removes unused websocket gateway and events client from the Origin.


e00dan (approved):

e00dan (changes requested):
Email notifications need to work after removing websocket events

e00dan (approved):

PR #837, opened 2020-04-11
fix(migrations): fix yarn migrate:demo command

e00dan:


kosecki123 (approved):
Ow, bug. Thanks for fixing.

josipbagaric (approved):

PR #842, opened 2020-04-13
Remove requirement for superuser in migration seed.sql

kosecki123:
This PR fixes the requirement of a DB superuser to run the seed.sql

```
SET session_replication_role = replica;
```

was replaced with insert and update sequence.


josipbagaric (approved):

josipbagaric:
Awesome!

PR #857, opened 2020-04-16
chore(origin-backend-app): print package version on start

kosecki123:


kosecki123:
@Kuzirashi this is only for the `origin-backend-app` how could we apply similar feature to the `origin-ui`?

josipbagaric:
Can it also print versions of the `exchange` and `origin-backend`?

kosecki123:
> Can it also print versions of the `exchange` and `origin-backend`?

Good idea. Added.

josipbagaric (approved):

PR #859, opened 2020-04-16
fix(origin-ui-core): solidify custom row display in exchange ask details

e00dan:
- Use less naive implementation and display ask details for order id, not order index in table
- Add support in TableMaterial to use rows string ids as TableRow keys (this allows React to more effectively manage rows when sorting etc.)
- Fix bug where ask details were staying open even when order was removed from the table for example due to filtering
- Use cleaner and more flexible implementation for displaying custom row in TableMaterial


kosecki123 (approved):
LGTM

PR #861, opened 2020-04-16
feat(origin-backend-app): enable swagger endpoint

kosecki123:
This PR aims to enable swagger site for both backend and exchange under origin-backend-app runner.

The issue with generating rich swagger site is not solved in this PR.


kosecki123:
This has one issue, it does not really take advantage of nest/swagger plugin which scans and generats rich swagger data.

josipbagaric (approved):

e00dan (approved):

PR #895, opened 2020-04-21
Fix: CertificationRequest backend data unavailable

josipbagaric:


kosecki123 (approved):

packages/origin-ui-core/src/features/general/sagas.ts
```diff
@@ -321,7 +321,7 @@ function* fetchDataAfterConfigurationChange(configuration: Configuration.Entity)
         [configuration]
     );
 
-    for (const certificate of certificates) {
+    for (const certificate of certificates.filter((cert) => cert.initialized)) {
```
e00dan: I think it's better to declare filter outside of for body. It might filter on every iteration otherwise (not sure)

josipbagaric: This works fine, but I will make the change to make your week happier :).

PR #908, opened 2020-04-22
chore(issuer): improved fetching certificates

### LINKED ISSUE #904: Not efficient certificates fetching
Currently the code responsible for fetching certificates uses total request and N events fetching events

```
export async function getAllCertificates(
    configuration: Configuration.Entity
): Promise<Certificate[]> {
    const { issuer } = configuration.blockchainProperties as Configuration.BlockchainProperties<
        Registry,
        Issuer
    >;
    const totalRequests = (await issuer.totalRequests()).toNumber();

    const certificatePromises = Array(totalRequests)
        .fill(null)
        .map(async (item, index) => {
            const certificationRequestApprovedEvents = await getEventsFromContract(
                issuer,
                issuer.filters.CertificationRequestApproved(null, index + 1, null)
            );

            if (certificationRequestApprovedEvents.length < 1) {
                return null;
            }

            const certId = certificationRequestApprovedEvents[0]._certificateId;
            return certId.gt(0) ? new Certificate(certId.toNumber(), configuration).sync() : null;
        });

    const certificates = await Promise.all(certificatePromises);

    return certificates.filter((cert) => cert !== null);
}
```

This could be replaced with one call to fetch the IDs based on approved events

```
const certificationRequestApprovedEvents = await getEventsFromContract(
                issuer,
                issuer.filters.CertificationRequestApproved(null, null, null)
            );
```

Then `new Certificate(id).sync()` can be used to return Certificate objects.

josipbagaric:
Closes https://github.com/energywebfoundation/origin/issues/904.

### DIFF
--- a/packages/issuer/src/test/Issuer.test.ts
+++ b/packages/issuer/src/test/Issuer.test.ts
@@ -9,7 +9,7 @@ import { OffChainDataSourceMock } from '@energyweb/origin-backend-client-mocks';
 
 import { providers, Wallet } from 'ethers';
 import { migrateIssuer, migrateRegistry } from '../migrate';
-import { CertificationRequest } from '..';
+import { CertificationRequest, getAllCertificationRequests } from '..';
 
 import { logger } from '../Logger';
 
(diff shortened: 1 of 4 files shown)


packages/issuer/src/test/Issuer.test.ts
```diff
@@ -66,6 +66,16 @@ describe('Issuer', () => {
         };
     });
 
+    it('gets all certification requests', async () => {
+        const totalVolume = new BigNumber(1e9);
+
+        await createCertificationRequest(totalVolume);
+        await createCertificationRequest(totalVolume);
```
kosecki123, line 73: Why duplicated?

kosecki123 (approved):
👍

packages/issuer/src/test/Issuer.test.ts
josipbagaric: So that it creates two certification requests. And then the tests checks if 2 were created.

e00dan (approved):

PR #914, opened 2020-04-24
chore(utils-general): remove privateKey from Configuration.blockchain…

josipbagaric:
…Properties


kosecki123:
Do we need that Configuration.Entity at all? This could be replaced with 3 parameters in the constructor instead.

kosecki123:
Shall we close this PR @JosephBagaric ?

PR #915, opened 2020-04-24
Bind CertificationRequests to Device

josipbagaric:
- [x] **Bind Devices and Certification Requests** - will be used to evaluate determine which smart meter readings have been certified
- [x] Add a `CertificationRequestWatcher` on the backend to **read and store incoming requests directly from on-chain**
- [x] Enable the certification request watcher to read approved and revoked events
- [x] Enable signing messages with a private key
- [x] Fix the UI integration
- [ ] Create a queue for updating CertificationRequests after they've been read from the blockchain event


packages/issuer/src/blockchain-facade/CertificationRequest.ts
```diff
@@ -91,10 +92,14 @@ export class CertificationRequest extends PreciseProofEntity implements ICertifi
             events.find((log: BlockchainEvent) => log.event === 'CertificationRequested').topics[2]
         );
 
-        const success = await configuration.offChainDataSource.certificateClient.updateCertificationRequestData(
-            request.id,
-            { energy, files }
-        );
+        const success = await polly()
```
kosecki123, line 95: What is the reason of polling here? Are we waiting for the backend to create the request based on the blockchain event?

josipbagaric: Yes

josipbagaric: Now I see that we can do it better. We can create a queue on the backend for these updates, so that when the backend reads the event, the queue is checked and entity updated.

packages/issuer/src/blockchain-facade/CertificationRequest.ts
```diff
@@ -108,51 +113,11 @@ export class CertificationRequest extends PreciseProofEntity implements ICertifi
...
-
-        const certificationRequestedLogs = await getEventsFromContract(
-            issuer,
-            issuer.filters.CertificationRequested(null, this.id, null)
-        );
-
-        const creationBlock = await issuer.provider.getBlock(
-            certificationRequestedLogs[0].blockNumber
+        const offChainData = await this.configuration.offChainDataSource.certificateClient.getCertificationRequest(
+            this.id
```
kosecki123, line 117: Is there a case for origin where this would be better read from the blockchain?

josipbagaric: Not sure. This used to be the case. Reading it from the backend is more secure, albeit more centralized.

packages/issuer/src/test/Issuer.test.ts
```diff
@@ -34,16 +34,45 @@ describe('Issuer', () => {
...
+        fromTime?: number,
+        toTime?: number,
+        deviceId?: string
+    ) => {
         setActiveUser(deviceOwnerWallet);
 
-        const fromTime = timestamp;
+        const generationFromTime = fromTime ?? timestamp;
         // Simulate time moving forward 1 month
         timestamp += 30 * 24 * 3600;
```
kosecki123, line 48: use moment library to calculate one calendar month properly. Ok, nevermind this is test.

kosecki123 (approved):

packages/issuer/src/blockchain-facade/CertificationRequest.ts
kosecki123: Reading from the blockchain would be required for on-chain issuance, like without external issuer. This good for now 👍

packages/issuer/src/blockchain-facade/CertificationRequest.ts
kosecki123: Yes, your proposal is ok, let's make it into separate PR. No need for that now.

PR #916, opened 2020-04-24
UI and API

m0azam:
Please review n comment


packages/localization/src/languages/pl.json
```diff
@@ -4,9 +4,14 @@
             "devices": "Urządzenia",
             "certificates": "Certyfikaty",
             "demands": "Oferty",
-            "organizations": "Organizacje"
+            "organizations": "Organizacje",
```
e00dan, line 7: It's okay but no need to add corresponding keys to other languages than `en.json` for now. If translation key is not present in other language than EN, it will fallback to EN. :)

packages/origin-backend-client/src/AdminClient.ts
```diff
@@ -0,0 +1,24 @@
+import { IUser } from '@energyweb/origin-backend-core';
+import { IRequestClient, RequestClient } from './RequestClient';
+
+type onUploadProgressFunction = (progressEvent: ProgressEvent) => void;
```
e00dan: Please remove this unused variable

packages/origin-backend-client/src/AdminClient.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface IAdminClient {
+    getAllUser(): Promise<IUser[]>;
```
e00dan: This is invalid English grammar. Please use: `getAllUsers` instead. 👍

packages/origin-backend-core/src/User.ts
```diff
@@ -9,13 +9,17 @@ type Currency = string;
...
 */
 export enum Role {
     UserAdmin,
     DeviceAdmin,
     DeviceManager,
     Trader,
-    Issuer
+    Issuer,
+    Admin,
+    Agents
```
e00dan: This should be singular form, so please use `Agent` instead of `Agents` as in other already existing roles. 👍

packages/origin-backend/src/pods/admin/admin.controller.ts
```diff
@@ -0,0 +1,14 @@
...
+import { AuthGuard } from '@nestjs/passport';
+import { AdminService } from './admin.service';
+
+@Controller('admin')
+export class AdminController {
+    constructor(private readonly adminService: AdminService) {}
+
+    @Get('users')
+    @UseGuards(AuthGuard('jwt'))
+    public async getAllUser() {
```
e00dan: Please use `getAllUsers` name instead of `getAllUser`.

kosecki123 (changes requested):
Before the review, rebase the commits and follow the conventional commit scheme.

packages/origin-backend-client/src/AdminClient.ts
m0azam: updated

PR #968, opened 2020-05-02
feat: admin user browsing

m0azam:
Please review this branch

### DIFF
--- a/packages/origin-backend-client-mocks/src/OffChainDataSourceMock.ts
+++ b/packages/origin-backend-client-mocks/src/OffChainDataSourceMock.ts
@@ -41,4 +41,6 @@ export class OffChainDataSourceMock implements IOffChainDataSource {
         this.deviceClient = new DeviceClientMock();
         this.organizationClient = new OrganizationClientMock();
     }
+
+    adminClient: import('../../origin-backend-client/src').IAdminClient;
 }
--- a/packages/origin-backend-client/src/AdminClient.ts
+++ b/packages/origin-backend-client/src/AdminClient.ts
@@ -1,0 +1,30 @@
+import { IUser, UserUpdateData, IUserWithRelations } from '@energyweb/origin-backend-core';
+import { IRequestClient, RequestClient } from './RequestClient';
+
+export interface IAdminClient {
+    update(formData: UserUpdateData): Promise<IUserWithRelations>;
+    getAllUsers(): Promise<IUser[]>;
+}
+
+export class AdminClient implements IAdminClient {
+    constructor(
+        private readonly dataApiUrl: string,
+        private readonly requestClient: IRequestClient = new RequestClient()
+    ) {}
+
+    async update(formData: IUser) {
+        console.log(formData);
+        const response = await this.requestClient.put(`${this.endpoint}/users/` + formData.id,formData);
+        console.log(response);
+        return response.data;
+    }
+
+    public async getAllUsers() {
+        const { data } = await this.requestClient.get(`${this.endpoint}/users`);
+        return data;
+    }
+
+    private get endpoint() {
+        return `${this.dataApiUrl}/admin`;
+    }
+}
--- a/packages/origin-backend-core/src/User.ts
+++ b/packages/origin-backend-core/src/User.ts
@@ -9,13 +9,30 @@ type Currency = string;
     DeviceManager:    0x0...----1-- = 4
     Trader:           0x0...---1--- = 8
     Issuer:           0x0...--1---- = 16
+    Admin:            0x0...-1----- = 32
+    Agents:           0x0...1------ = 64
 */
 export enum Role {
     UserAdmin,
     DeviceAdmin,
     DeviceManager,
     Trader,
-    Issuer
+    Issuer,
+    Admin,
+    Agent
+}
+
+export enum Status {
+    'Pending',
+    'Active',
+    'Suspended',
+    'Deleted'
+}
+
+export enum KYCStatus {
+    'Pending KYC',
+    'KYC passed',
+    'KYC rejected'
 }
 
 export function buildRights(roles: Role[]): number {
@@ -56,6 +73,8 @@ export interface IUserProperties {
     notifications: boolean;
     autoPublish: IAutoPublishConfig;
     rights: number;
+    status: number;
+    kycStatus: number;
 }
 
 export interface IUser extends IUserProperties {
--- a/packages/origin-backend/migrations/initial.sql
+++ b/packages/origin-backend/migrations/initial.sql
@@ -163,7 +163,9 @@ CREATE TABLE public."user" (
     notifications boolean,
     "autoPublish" text,
     rights integer DEFAULT 0 NOT NULL,
-    "organizationId" integer
+    "organizationId" integer,
+    "status" integer DEFAULT 0 NOT NULL,
+    "kycStatus" integer DEFAULT 0 NOT NULL
 );
 
 CREATE SEQUENCE public.user_id_seq
--- a/packages/origin-backend/src/pods/admin/admin.controller.ts
+++ b/packages/origin-backend/src/pods/admin/admin.controller.ts
@@ -1,0 +1,35 @@
+import { IUser, SupportedEvents, UserStatusChangedEvent } from '@energyweb/origin-backend-core';
+import { Body, Controller, Get, Param, Put, UseGuards } from '@nestjs/common';
+import { AuthGuard } from '@nestjs/passport';
+import { NotificationService } from '../notification/notification.service';
+import { AdminService } from './admin.service';
+
+@Controller('admin')
+export class AdminController {
+    constructor(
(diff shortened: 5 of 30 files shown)


packages/origin-backend-client-mocks/src/OffChainDataSourceMock.ts
```diff
@@ -41,4 +41,6 @@ export class OffChainDataSourceMock implements IOffChainDataSource {
         this.deviceClient = new DeviceClientMock();
         this.organizationClient = new OrganizationClientMock();
     }
+
+    adminClient: import('../../origin-backend-client/src').IAdminClient;
```
kosecki123, line 45: This should be imported from '@energyweb/origin-backend-client'

packages/origin-backend-client/src/AdminClient.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export class AdminClient implements IAdminClient {
+    constructor(
+        private readonly dataApiUrl: string,
+        private readonly requestClient: IRequestClient = new RequestClient()
+    ) {}
+
+    async update(formData: IUser) {
+        console.log(formData);
```
kosecki123, line 16: Rm console.log

packages/origin-backend-client/src/AdminClient.ts
```diff
@@ -0,0 +1,30 @@
...
+        const response = await this.requestClient.put(`${this.endpoint}/users/` + formData.id,formData);
+        console.log(response);
```
kosecki123, line 18: Rm console.log

packages/origin-backend/migrations/initial.sql
```diff
@@ -163,7 +163,9 @@ CREATE TABLE public."user" (
     notifications boolean,
     "autoPublish" text,
     rights integer DEFAULT 0 NOT NULL,
-    "organizationId" integer
+    "organizationId" integer,
+    "status" integer DEFAULT 0 NOT NULL,
```
kosecki123, line 167: This should be removed, initial.sql is not meant to be edited, all changes are handled using migrations/* files

packages/origin-backend-core/src/User.ts
```diff
@@ -9,13 +9,30 @@ type Currency = string;
...
 */
 export enum Role {
     UserAdmin,
     DeviceAdmin,
     DeviceManager,
     Trader,
-    Issuer
+    Issuer,
+    Admin,
+    Agent
```
kosecki123, line 22: I suggest we rename this to SupportAgent.

packages/origin-backend/src/pods/admin/admin.controller.ts
```diff
@@ -0,0 +1,35 @@
...
+export class AdminController {
+    constructor(
+        private readonly notificationService: NotificationService,
+        private readonly adminService: AdminService
+    ) {}
+
+    @Get('users')
+    @UseGuards(AuthGuard('jwt'))
+    public async getAllUsers() {
+        return this.adminService.getAllUsers();
```
kosecki123, line 17: Since we know that this endpoint can be only used by Admin and Agent (SupportAgent) roles, please add the authorization.

An example snippet:

```
 @Get('/:id/devices')
    @UseGuards(AuthGuard('jwt'))
    async getDevices(@Param('id') id: string, @UserDecorator() loggedUser: IUserWithRelationsIds) {
        if (!isRole(loggedUser, Role.DeviceManager)) {
            throw new ForbiddenException();
        }
```

packages/origin-backend/src/pods/admin/admin.controller.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Put('users/:id')
+    @UseGuards(AuthGuard('jwt'))
+    public async put(@Param('id') id: string, @Body() body: IUser) {
```
kosecki123, line 22: Same as above, add authorization.

packages/origin-backend/src/pods/user/user.entity.ts
```diff
@@ -49,4 +49,10 @@ export class User extends ExtendedBaseEntity implements IUser {
 
     @Column({ default: 0, nullable: false })
     rights: number;
+
+    @Column({ default: 0, nullable: false })
+    status: number;
```
kosecki123, line 54: Use Status enum instead number.

packages/origin-backend/src/pods/user/user.entity.ts
```diff
@@ -49,4 +49,10 @@ export class User extends ExtendedBaseEntity implements IUser {
...
+
+    @Column({ default: 0, nullable: false })
+    kycStatus: number;
```
kosecki123, line 57: Use KycStatus enum instead.

packages/origin-backend/src/pods/admin/admin.service.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    async update(id: number | string, data: IUser): Promise<ExtendedBaseEntity & IUser> {
+        const entity = await this.repository.findOne(id);
+
+        if (!entity) {
+            throw new Error(`Can't find entity.`);
+        }
+
+        await this.repository.update(id, {
```
kosecki123, line 28: Data you are storing, does not seems to be validated at all. Looks like partial validation is implemented using annotation in the `export class User extends ExtendedBaseEntity implements IUser`.

While this is not the best approach, please follow this path similar to:

```
async create(data: DeviceCreateData & Pick<IDeviceWithRelationsIds, 'organization'>) {
        const configuration = await this.configurationService.get();

        const newEntity = new Device();

        Object.assign(newEntity, {
            ...data,
            externalDeviceIds: data.externalDeviceIds
                ? data.externalDeviceIds.map(({ id, type }) => {
                      if (
                          typeof id === 'undefined' &&
                          configuration.externalDeviceIdTypes?.find((t) => t.type === type)
                              ?.autogenerated
                      ) {
                          return { id: uuid(), type };
                      }

                      return { id, type };
                  })
                : []
        });

        const validationErrors = await validate(newEntity);

        if (validationErrors.length > 0) {
            throw new UnprocessableEntityException({
                success: false,
                errors: validationErrors
            });
        }

        await this.repository.save(newEntity);

        return newEntity;
    }
```

Create empty entity, apply changes, run validation, check results...

packages/origin-backend/src/pods/notification/notification.service.ts
```diff
@@ -86,6 +90,13 @@ export class NotificationService {
                     DeviceStatus[data.status]
                 }".<br /><br /><a href="${url}">${url}</a>`
             );
+        },
+        [SupportedEvents.USER_STATUS_CHANGED]: async (data: UserStatusChangedEvent) => {
+            await this.sendNotificationEmail(
+                EmailTypes.USER_STATUS_CHANGED,
+                data.email,
+                `Status of your user information`
```
kosecki123, line 98: Replace with `Your status has been changed.`

kosecki123 (changes requested):

packages/origin-ui-core/src/components/AdminUsersTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+    usePaginatedLoader
+} from './Table/PaginatedLoaderHooks';
+import { TableMaterial } from './Table/TableMaterial';
+import { useHistory } from 'react-router-dom';
+
+interface IRecord {
+    user: IUser;
+}
+
+export const Status = {
```
josipbagaric, line 18: Duplicate declaration. This enum can be imported from the `origin-backend-core` package.

packages/origin-ui-core/src/components/AdminUsersTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+    1: 'Pending',
+    2: 'Active',
+    3: 'Suspended',
+    4: 'Deleted'
+};
+
+export const KYCStatus = {
```
josipbagaric, line 25: Duplicate declaration. This enum can be imported from the `origin-backend-core` package.

packages/origin-ui-core/src/components/AdminUsersTable.tsx
m0azam: This are based on object key-value

packages/origin-ui-core/src/components/AdminUsersTable.tsx
m0azam: This are based on object key-value

m0azam: i rename to KYCStatus

packages/origin-ui-core/src/components/AdminUsersTable.tsx
m0azam: rename to KeyStatus

kosecki123:
Duplicate of #973

PR #979, opened 2020-05-07
chore(origin-ui-core):Admin panel, when clicking row show user details

m0azam:


packages/origin-backend-client/src/AdminClient.ts
```diff
@@ -21,6 +22,14 @@ export class AdminClient implements IAdminClient {
         const { data } = await this.requestClient.get(`${this.endpoint}/users`);
         return data;
     }
+    
+    public async getUsersBy(orgName: string,status: number, kycStatus: number) {
+        orgName = orgName == null?'':orgName;
```
kosecki123: Looks like origin-backend-client does not have linters/prettier enabled. 

I'd rather have this:

const organizationName = orgName ?? '';

packages/origin-backend-client/src/AdminClient.ts
```diff
@@ -21,6 +22,14 @@ export class AdminClient implements IAdminClient {
         const { data } = await this.requestClient.get(`${this.endpoint}/users`);
         return data;
     }
+    
+    public async getUsersBy(orgName: string,status: number, kycStatus: number) {
```
kosecki123: Use enums instead of numbers here, it would make this more readable.

packages/origin-backend-client/src/AdminClient.ts
```diff
@@ -21,6 +22,14 @@ export class AdminClient implements IAdminClient {
...
+        orgName = orgName == null?'':orgName;
+        status = isNaN(status)?0:status;
+        kycStatus = isNaN(kycStatus)?0:kycStatus;
+        const { data } = await this.requestClient.get(`${this.endpoint}/usersBy?orgName=` + orgName + `&status=` + status + `&kycStatus=` + kycStatus );
```
kosecki123: Consider using string interpolation

`orgName=${orgName)&....`

packages/origin-backend/src/pods/admin/admin.controller.ts
```diff
@@ -25,6 +25,16 @@ export class AdminController {
         return this.adminService.getAllUsers();
     }
 
+    @Get('usersBy')
+    @UseGuards(AuthGuard('jwt'))
```
kosecki123, line 31: Missing Roles annotations.

packages/origin-backend/src/pods/admin/admin.service.ts
```diff
@@ -19,6 +19,17 @@ export class AdminService {
         });
     }
 
+    public async getUsersBy(orgName: string, status: number, kycStatus: number) {
+        const result = await this.repository.find({
+            where: { status, kycStatus },
+            relations: ['organization']
+        });
+        return result.filter((item) => {
```
kosecki123: Read this on how to handle case insensitive queries https://github.com/typeorm/typeorm/issues/1231

kosecki123 (changes requested):

packages/origin-backend-client/src/AdminClient.ts
josipbagaric: Can be simplified to `orgName ?? ''`

josipbagaric (approved):
LGTM.

packages/origin-backend/src/pods/admin/admin.service.ts
m0azam: have do something as below , but not working . any other sample that can help on join

public async getUsersBy(orgName: string, status: Status, kycStatus: KYCStatus) {
        console.log({ name:Like(`%${orgName}%`) });
        const result = await this.repository.find({
            where: { status, kycStatus, organization:{ name: Raw(name => `LOWER(${name}) Like '%${orgName.toLowerCase()}%`)}  },
            relations: ['organization']
        });
        console.log(result);
        return result;

josipbagaric:
@m0azam Can I ask you not to force-push commits after you've requested a review on a PR?
Especially after the reviews have been done.

It really makes it hard for us to review the PR and track the changes, because the history is overwritten and then we have to re-review the whole PR again instead of individual new commits.

m0azam:
> @m0azam Can I ask you not to force-push commits after you've requested a review on a PR?
> Especially after the reviews have been done.
> 
> It really makes it hard for us to review the PR and track the changes, because the history is overwritten and then we have to re-review the whole PR again instead of individual new commits.

ok will do it after this, sorry on this issue

packages/origin-backend/src/pods/admin/admin.controller.ts
```diff
@@ -25,6 +27,16 @@ export class AdminController {
...
+    public async getUsersBy(
```
kosecki123, line 33: Still missing roles.

kosecki123 (changes requested):
LGTM, just add the roles in the controller.

kosecki123 (approved):

PR #980, opened 2020-05-07
chore(origin-backend): update migrations

josipbagaric:
Adds missing migrations.


packages/origin-backend/migrations/1588836124586-UserStatusAndAutoPublish.ts
```diff
@@ -0,0 +1,26 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class Status1588836124586 implements MigrationInterface {
+    name = 'Status1588836124586';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "user" DROP COLUMN "autoPublish"`, undefined);
```
kosecki123: 👍

packages/origin-backend/migrations/1588836124586-UserStatusAndAutoPublish.ts
```diff
@@ -0,0 +1,26 @@
...
+        await queryRunner.query(`ALTER TABLE "user" ALTER COLUMN "status" DROP DEFAULT`, undefined);
```
kosecki123: There should be default set. The fix should add that default option in to entity.ts file, so the typeorm migration tool will see that.

kosecki123 (changes requested):

packages/origin-backend/src/pods/user/user.entity.ts
```diff
@@ -47,9 +47,9 @@ export class User extends ExtendedBaseEntity implements IUser {
     @Column({ default: 0, nullable: false })
     rights: number;
 
-    @Column()
+    @Column({ default: 0 })
```
kosecki123: Can we use the enum instead of 0?

kosecki123 (approved):

packages/origin-backend/src/pods/user/user.entity.ts
josipbagaric: I think we might, just will need to rename the Enums because some use spaces in their names which causes syntax errors.

m0azam: for Column can add { default: 0, nullable: false }

PR #982, opened 2020-05-07
fix(migrations): remove autopublish from seed.sql

josipbagaric:
Fixes an issue where migrations wouldn't run because `autoPublish` was still set in seed.sql.


kosecki123 (approved):
😅

PR #1004, opened 2020-05-11
chore(origin-ui-core):Add upload component in form

m0azam:


josipbagaric (approved):
LGTM. Would be cool to have screenshots attached to the PR description when a UI change happened.

packages/origin-ui-core/src/components/DeviceGroupForm.tsx
```diff
@@ -122,6 +123,17 @@ export function DeviceGroupForm(props: IProps) {
 
     const classes = useStyles(useTheme());
 
+    const [files, setFiles] = useState<IUploadedFile[]>([]);
+    const uploadedFiles = files
+        .filter((f) => !f.removed && f.uploadedName)
+        .reduce(
```
kosecki123, line 129: Perhaps easier and faster way to do this is to use

`.map(x=>x.uploadedName)` 

instead of `.reduce`

kosecki123 (approved):

PR #1015, opened 2020-05-13
fix(origin-backend): device files migrations

josipbagaric:


m0azam (commented):
already have in this file 1589264954930-UploadFilesInDevice.ts

josipbagaric:
> already have in this file 1589264954930-UploadFilesInDevice.ts

This new migration was autogenerated by TypeORM. Means something is wrong.

josipbagaric:
Added a default value of `''`. This seems to be working fine for me.

m0azam (approved):

PR #1016, opened 2020-05-13
fix: fixate tslib version to 1.11.2 to avoid typeorm issue

josipbagaric:


kosecki123:
It’s good but missing the fix for the Dockerfile in migrations/canary 

There is a fixed version 0.2.22 now so it requires the installation of this particular tslib

kosecki123 (approved):

PR #1017, opened 2020-05-13
Fix invalid type when query for Certification Request record

atchapcyp:
https://github.com/energywebfoundation/origin/blob/d12e551b12e8832b0d31dc9d3f8dd835cad5c6d9/packages/origin-backend/src/pods/certificate/certification-request.service.ts#L44

https://github.com/energywebfoundation/origin/blob/d12e551b12e8832b0d31dc9d3f8dd835cad5c6d9/packages/origin-backend/src/pods/certificate/certification-request-watcher.service.ts#L151

Since`registerApproved` take an arguement as `number`, We should convert them as type number instead of BigNumber.


josipbagaric:
Looks good. Please run yarn lint-fix to fix lint errors.

josipbagaric (approved):

PR #1056, opened 2020-05-18
Fix account data return values

kosecki123:
This PR provides:

- Fix for an issue where `locked` assets were returned with negative sign
- `Available` and `locked` with amount equal 0 are not returned as account assets


mariopavlovic (approved):
Looks good

PR #1058, opened 2020-05-20
Request certificates with organization deposit account as default

kosecki123:
This PR changes the way certificates requests are being handled. 

Before:
1) device manager requested certificate
2) issuer approves
3) certificate issued to the `sender` account (device manager in this case)
4) device manager transfer certificates to the exchange

Current:
1) device manager requested certificate
2) issuer approves
3) certificate issued to the organization deposit account on exchange

### DIFF
--- a/packages/issuer/src/blockchain-facade/CertificationRequest.ts
+++ b/packages/issuer/src/blockchain-facade/CertificationRequest.ts
@@ -79,14 +79,15 @@ export class CertificationRequest extends PreciseProofEntity implements ICertifi
             ? issuerWithSigner.requestCertificationFor(data, forAddress, isVolumePrivate)
             : issuerWithSigner.requestCertification(data, isVolumePrivate));
 
-        const { events } = await tx.wait();
+        const {
+            events: [certificationRequested]
+        } = await tx.wait();
 
-        request.id = Number(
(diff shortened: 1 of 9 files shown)


packages/issuer/src/blockchain-facade/CertificationRequest.ts
```diff
@@ -79,14 +79,15 @@ export class CertificationRequest extends PreciseProofEntity implements ICertifi
...
 
-        const { events } = await tx.wait();
+        const {
+            events: [certificationRequested]
+        } = await tx.wait();
 
-        request.id = Number(
-            events.find((log: BlockchainEvent) => log.event === 'CertificationRequested').topics[2]
-        );
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mariopavlovic, line 86: not clear why we have this comment here

mariopavlovic (approved):
Looks good

packages/issuer/src/blockchain-facade/CertificationRequest.ts
kosecki123: Using `any` is by default treated as a error by linter, which is in 99.9% cases correct, this is an exemption and this comment make linter not throwing an error.

PR #1079, opened 2020-05-25
feat(exchange): add publishing trade event to messaging queue for external services

chonla:
This PR adds publishing trade event to messaging queue for external services. The published item is trade item in TradeExecutedEvent.

The feature is feature toggle-able by setting `USE_MESSAGE_QUEUE` in .env file, default is `false`. Other configuration, like queue connection string, queue name, and queue topic, please see in `.env.example`.


poorprogrammer:
I personally like this feature, more friendly with Domain-Driven Design and extendable.

PR #1084, opened 2020-05-26
Revert "fix(deps): update dependency @date-io/moment to v2"

josipbagaric:
Reverts energywebfoundation/origin#427 because it causes issues with **@material-ui/pickers v3.** Can be revisited after **@material-ui/pickers v4** reaches stable.

Screenshot 2020-05-26 at 14 56 51


kosecki123 (approved):

kosecki123:
@JosephBagaric could you also add this to ignore list on Renovate? I recently changed the settings so this will get auto updated next weekend.

PR #1088, opened 2020-05-27
fix: endpoint for fetching invitations for an organization

josipbagaric:
Fixes fetching invitations for an organization.

### DIFF
--- a/packages/origin-backend/src/pods/organization/organization.controller.ts
+++ b/packages/origin-backend/src/pods/organization/organization.controller.ts
@@ -190,6 +190,25 @@ export class OrganizationController {
         };
     }
 
+    @Get('/invitation/:organizationId')
+    @UseGuards(AuthGuard())
+    async getInvitationsForOrganization(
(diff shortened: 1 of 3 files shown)


packages/origin-backend/src/pods/organization/organization.controller.ts
```diff
@@ -190,6 +190,25 @@ export class OrganizationController {
         };
     }
 
+    @Get('/:id/invitations')
+    @UseGuards(AuthGuard())
+    async getInvitationsForOrganization(
+        @Param('id') organizationId: string
+    ): Promise<IOrganizationInvitation[]> {
+        const organization = await this.organizationService.findOne(organizationId);
```
kosecki123, line 206: This allows any logged in user to browse every organization invitations. I think this should not be possible. 

You can use UserDecorator and read the organization id of the logged in user.

kosecki123 (changes requested):

kosecki123 (approved):

PR #1089, opened 2020-05-28
Feat/profile page

m0azam:
New User profile page

### DIFF
--- a/packages/origin-backend/src/pods/user/user.controller.ts
+++ b/packages/origin-backend/src/pods/user/user.controller.ts
@@ -3,7 +3,9 @@ import {
     IUserWithRelationsIds,
     UserRegisterReturnData,
     UserRegistrationData,
-    UserUpdateData
+    UserUpdateData,
+    IUser,
+    UserPasswordUpdate
 } from '@energyweb/origin-backend-core';
 import { UserDecorator } from '@energyweb/origin-backend-utils';
 import {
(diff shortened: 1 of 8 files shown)


packages/origin-backend/src/pods/user/user.controller.ts
```diff
@@ -73,4 +75,22 @@ export class UserController {
...
+
+    @Put('profile/:id')
+    @UseGuards(AuthGuard('jwt'))
+    public async putProfile(@Param('id') id: string, @Body() body: IUser) {
+        return this.userService.updateProfile(id, body);
+    }
+
+    @Put('password')
+    @UseGuards(AuthGuard('jwt'))
+    public async putChainAddress(@Body() body: UserPasswordUpdate) {
```
josipbagaric: The user should only be able to update his own password (or perhaps OrganizationAdmin can also do that?).
Anyway, this requires proper auth.
You can use `@UserDecorator` for this.

Also, wondering why the method is called `putChainAddress()` but it updates the user password.

packages/origin-backend/src/pods/user/user.controller.ts
```diff
@@ -73,4 +75,22 @@ export class UserController {
...
+        return this.userService.updatePassword(body);
+    }
+
+    @Put('chainAddress/:id')
+    @UseGuards(AuthGuard('jwt'))
+    public async putPassword(@Param('id') id: string, @Body() body: IUser) {
+        return this.userService.updateBlockChainAddress(id, body);
```
josipbagaric, line 97: The method is called `putPassword()` but updates blockchain address.

Also missing proper auth. The user should only be able to update his own blockchain account address.

josipbagaric (changes requested):

josipbagaric (approved):

PR #1090, opened 2020-05-28
feat(origin-ui-core): new ui for automatic supply price for devices

m0azam:


packages/origin-backend/src/pods/device/device.controller.ts
```diff
@@ -110,7 +113,7 @@ export class DeviceController {
 
         if (
             body?.automaticPostForSale === undefined ||
-            (body?.automaticPostForSale &&
+            (body?.automaticPostForSale === undefined &&
```
m0azam: add this because variable are boolean

josipbagaric:
@m0azam Looks good to me, but tests need to be fixed.

packages/origin-backend-client-mocks/src/DeviceClientMock.ts
```diff
@@ -77,4 +70,17 @@ export class DeviceClientMock implements IDeviceClient {
...
         this.storage.set(id, device);
     }
+
+    public async getSupplyBy(facilityName: string, status: number): Promise<IDeviceWithRelationsIds[]> {
+        return [...this.storage.values()];
+    }
+
+    public async delete(id: number): Promise<void> {
+        
+    }
```
josipbagaric, line 80: There were some `throw()` methods here. What's the reason for removing them and this being empty now?

m0azam:
Screen Shot 2020-05-29 at 10 11 34 PM

packages/origin-backend-client-mocks/src/DeviceClientMock.ts
m0azam: do i need put back as though that are reason of test failed

josipbagaric (approved):

PR #1092, opened 2020-05-28
Transition to event driven architecture

kosecki123:
This PR changes the way trades that occur on the matching engine are communicated outside. This allows 3rd party developers to write custom event handlers on runner application layer for e.g `@energyweb/origin-backend-app`

Closes #1079


josipbagaric (approved):
LGTM

PR #1094, opened 2020-05-29
Change Organization Member Role

josipbagaric:
Adds a function to change roles for organization users from the Origin UI.

Screenshot 2020-06-01 at 13 39 30

- [x] Remove reference to (currently unused) LeadUser
- [x] Show user roles on the Organization Users table
- [x] Backend for changing a role for an organization member. There always has to be at least one OrganizationAdmin.
- [x] UI selector modal for selecting the new role
- [x] Simplified Role `<Select />` element when inviting users
- [x] Refresh current user role automatically after updating
- [ ] ~(Optional) Add i18n to emails (backend)~ Will create a new PR


packages/origin-backend-core/src/User.ts
```diff
@@ -32,6 +32,30 @@ export function buildRights(roles: Role[]): number {
...
+    for (const role of roles.sort((a, b) => b - a)) {
+        if (userRights < role) {
+            continue;
+        }
+
+        userRoles.push(role);
+        userRights -= role;
+    }
+
+    return userRoles;
```
kosecki123: More elegant solution:

```
return roles.filter(role => rights&role);
```

kosecki123 (approved):

PR #1113, opened 2020-06-01
Use mocha instead of jest for e2e tests

kosecki123:
This PR changes the e2e runner in exchange and origin-backend from jest to mocha. Seems like `ts-jest + jest` is less reliable and less performant than `ts-node + mocha`. 

Also since #1112 ts-node can take advantage of composite builds.


packages/exchange/src/pods/account-balance/account-balance.service.spec.ts
```diff
@@ -17,10 +19,10 @@ import { Transfer } from '../transfer/transfer.entity';
...
 import { AccountBalanceService } from './account-balance.service';
 
-jest.mock('../trade/trade.service');
-jest.mock('../transfer/transfer.service');
-jest.mock('../order/order.service');
-jest.mock('../bundle/bundle.service');
+// jest.mock('../trade/trade.service');
+// jest.mock('../transfer/transfer.service');
+// jest.mock('../order/order.service');
+// jest.mock('../bundle/bundle.service');
```
josipbagaric: I guess this can be removed.

josipbagaric (approved):

PR #1114, opened 2020-06-02
Fix/supply list by org and price with 2 decimal

m0azam:


packages/origin-backend/src/pods/device/device.controller.ts
```diff
@@ -46,11 +46,26 @@ export class DeviceController {
         return this.deviceService.getAll();
     }
 
+    @Get('/myDevice')
+    @UseGuards(AuthGuard(), RolesGuard)
+    @Roles(Role.OrganizationAdmin, Role.OrganizationDeviceManager, Role.OrganizationUser)
+    async getMyDevice(@UserDecorator() { organizationId }: ILoggedInUser) {
```
josipbagaric: The method is called `getMyDevice` but it fetches device**s**. Please change this to `getMyDevices()` to avoid confusion.

packages/origin-backend/src/pods/device/device.controller.ts
```diff
@@ -46,11 +46,26 @@ export class DeviceController {
         return this.deviceService.getAll();
     }
 
+    @Get('/myDevice')
```
josipbagaric: We don't use camelCase for URLs, but we use dash-separated URLs. Let's use `/my-devices` instead of `/myDevice`.

packages/origin-backend/src/pods/device/device.service.ts
```diff
@@ -185,6 +185,24 @@ export class DeviceService {
         return devices;
     }
 
+    async getMyDevice(organizationId: number, options: FindOneOptions<Device> = {}) {
```
josipbagaric: You don't need to create a new duplicate method for this, you can just use `getAll()`

```getAll({ where: { organization: { id: organizationId } } });```

**Review by josipbagaric — CHANGES_REQUESTED** [review]

File: packages/origin-ui-core/src/components/AutoSupplyDeviceTable.tsx [hunk 4]
```diff
@@ -80,19 +81,27 @@ export function AutoSupplyDeviceTable() {
...
+
     const columns = [
         { id: 'type', label: 'Type' },
         { id: 'facility', label: 'Facility' },
         { id: 'price', label: 'Price' },
         { id: 'status', label: 'Status' },
-        { id: 'certified', label: 'To be certified for 2020/2021 (Mwh)' }
+        {
+            id: 'certified',
+            label: `To be certified for ${currentYear.format('YYYY')}/${nextYear.format(
```
josipbagaric: Please use localized strings like this: https://github.com/energywebfoundation/origin/blob/ce900daed490d632ebee2feb6a4f2e4074ba149d/packages/origin-ui-core/src/components/ProducingDeviceDetailView.tsx#L141

For all these columns above.

josipbagaric (approved):

PR #1117, opened 2020-06-03
chore: improved direct buy

josipbagaric:
Wait for status before showing Success when direct buying.


kosecki123:
OrderStatus was in the past a part of exchange-core but since it's not longer in use there it was moved as enum to exchange.

The only reason to it was moved in this PR seems to be to be able to share it with the UI. On the other hand UI has it's own ExchangeClient with corresponding types https://github.com/energywebfoundation/origin/blob/c0113a1e72fabdf948b0a562beb752cc62a0eba4/packages/origin-ui-core/src/utils/exchange/ExchangeClient.ts#L36

Currently the clients code we use for API is manually created, this should ideally be done by Swagger client generators.

josipbagaric:
@kosecki123 Yes, the UI ExchangeClient does have it's own types, but they are still imported from `@energyweb/exchange-core` such as the `OrderSide` property: https://github.com/energywebfoundation/origin/blob/b2026f401cecbbef86f92222f07e0ad5269fba03/packages/origin-ui-core/src/utils/exchange/types.ts#L128

The only viable alternative would be to, as you said, client generators but they are outside of the scope of this PR.

JGiter (approved):

kosecki123 (approved):

PR #1120, opened 2020-06-03
bug: Repeat certificates in inbox

JGiter:
- [x] The certificate to be updated is determined both by identifier and by source
- [x] Enable redux-store update trace in redux-dev-tools


kosecki123 (approved):

josipbagaric (approved):

kosecki123:
Please rename the commit from bug()... to fix(...

PR #1121, opened 2020-06-03
Faster certificates inbox loading

kosecki123:
This PR changes the way certificates inbox is loaded.

Previously process required to all existing certificates to be synced, which over time was not sustainable.

Before:
- sync all existing certificates
- load owned exchange certificates
- filter by ower based on current metamask account

After:
- sync certificates which are currently owned
   - get all SingleTransfer events with "to" set to active address
   - get unique certificate id
   - get balances for certificate ids
   - sync certificates with non-zero balance
- load owned exchange certificates

### DIFF
--- a/packages/issuer/src/blockchain-facade/CertificateUtils.ts
+++ b/packages/issuer/src/blockchain-facade/CertificateUtils.ts
@@ -170,6 +170,33 @@ export async function getAllCertificates(
     return Promise.all(certificatePromises);
 }
 
+export async function getAllOwnedCertificates(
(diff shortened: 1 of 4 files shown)


josipbagaric (approved):

josipbagaric:
Works fine when switching users?

kosecki123:
> Works fine when switching users?

Blockchain users or logged in users?

For blockchain users, which is in-fact more of a demo/testing feature, it does not reload the inbox, but this is a *know issue* I guess.

PR #1123, opened 2020-06-04
fix: Certification Request off-chain data queue

josipbagaric:
### Previous behaviour
1. User creates a certification request on-chain
2. The backend reads the on-chain request and stores it in the database
3. User updates the backend certificate with off-chain data

This behaviour creates an issue where there would be an **invalid on-chain certification request** when the user would exit the application **after step 2 and before step 3**.

### New behaviour
1. User stores all off-chain certification request data in the backend
2. User creates a certificate on-chain
3. Backend reads the on-chain certificate and **updates it with all the off-chain data that are stored in the queue**.


packages/origin-backend/src/pods/certificate/certification-request-queue-item.entity.ts
```diff
@@ -0,0 +1,29 @@
...
+import { CertificationRequestQueueItemDTO } from './certification-request-queue-item.dto';
+
+@Entity()
+@Unique(['deviceId', 'fromTime', 'toTime'])
+export class CertificationRequestQueueItem extends ExtendedBaseEntity
+    implements CertificationRequestQueueItemDTO {
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Column('varchar', { nullable: true })
```
kosecki123: This is conflicting with 

```
@Validate(PositiveBNStringValidator)
    public readonly energy: string;
```

in DTO

kosecki123 (approved):

packages/origin-backend/src/pods/certificate/certification-request-queue-item.entity.ts
josipbagaric: You're correct, this shouldn't be nullable.

Issue #1081, opened 2020-05-26
Global ENERGY_PER_UNIT usage


josipbagaric:
https://github.com/energywebfoundation/origin/pull/1080 introduced a global variable called `ENERGY_PER_UNIT` which defines the amount of energy in single certificate in Wh for e.g 1000000 for 1MWh certificates.

This variable is currently used only in the `@energyweb/exchange` package, but should be used globally to avoid unknown errors and discrepancies.

PR #1140, opened 2020-06-08
Feature/withdraw from exchange

JGiter:
- add withdraw action in CertificateTable
- fix amount checking after withdrawal
- fix comparing withdrawal logs


packages/exchange/src/pods/withdrawal-processor/withdrawal-processor.service.ts
```diff
@@ -159,7 +145,10 @@ export class WithdrawalProcessorService implements OnModuleInit {
         this.logger.debug(`Withdrawal ${id} receipt: ${JSON.stringify(receipt)} `);
 
         const hasLog = receipt.logs
-            .map((log) => this.tokenInterface.parseLog(log))
+            .map((log) => {
+                const parsed = this.tokenInterface.parseLog(log);
```
kosecki123: This change is not necessary I guess.

packages/exchange/src/pods/withdrawal-processor/withdrawal-processor.service.ts
```diff
@@ -174,11 +163,14 @@ export class WithdrawalProcessorService implements OnModuleInit {
     }
 
     private hasMatchingLog(withdrawal: Transfer, log: ethers.utils.LogDescription) {
+        const _to = String(log.values._to).toLowerCase();
+        const _from = String(log.values._from).toLowerCase();
+        console.groupEnd();
```
kosecki123: Rm console.*

packages/localization/src/languages/pl.json
```diff
@@ -77,7 +77,8 @@
                 "claim": "Claim",
                 "returnToInbox": "Zwróć do skrzynki",
                 "buy": "Kup",
-                "request": "Żądaj"
+                "request": "Żądaj",
+                "withraw": "Wycofać"
```
kosecki123: Typo, should be `withdraw`

packages/origin-ui-core/src/components/Modal/WithdrawModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { EnergyFormatter, formatDate } from '../../utils';
+
+interface IProps {
+    certificate: ICertificateViewItem;
+    producingDevice: ProducingDevice.Entity;
+    showModal: boolean;
+    callback: () => void;
+}
+
+const DEFAULT_ENERGY_IN_BASE_UNIT = bigNumberify(1);
```
kosecki123: This should be changed into a .env variable as mentioned here https://github.com/energywebfoundation/origin/issues/1081

packages/origin-ui-core/src/features/general/sagas.ts
```diff
@@ -363,13 +363,14 @@ function* fetchDataAfterConfigurationChange(
     for (const device of producingDevices) {
         yield put(producingDeviceCreatedOrUpdated(device));
     }
+    console.group('>>> fetching after configuration changed <<<');
```
kosecki123: Rm console.* unless they are intended to stay.

kosecki123:
Does this allow user to withdraw a partial amount of the certificates?

packages/origin-ui-core/src/components/Modal/WithdrawModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+            requestWithdrawCertificate({
+                assetId,
+                address,
+                amount,
+                callback
+            })
+        );
+    }
+
+    const certificateId = certificate ? certificate.id : '';
```
kosecki123, line 104: In general not sure why this code is necessary, in the case that `certificate` object is incorrect you should return and add show error or something.

JGiter: the reason for this is that all component variables must be initialized regardless of whether the component is displayed (and the certificate is selected) or not

.env.example
```diff
@@ -50,4 +50,5 @@ ISSUER_ID=Issuer ID
 # Possible options: GRID_OPERATOR and LOCATION, should be comma separated
 DEVICE_PROPERTIES_ENABLED=GRID_OPERATOR,LOCATION
 # Amount of energy in single certificate 
-ENERGY_PER_UNIT=1000000
\ No newline at end of file
+ENERGY_PER_UNIT=1000000
+DEFAULT_ENERGY_IN_BASE_UNIT=1;
```
kosecki123: Why not use `ENERGY_PER_UNIT=1000000`? 1000000Wh is 1MWh which 1 from this newly added variable. Note that all operations in the system are denominated in Wh.

@JosephBagaric what's your opinion?

packages/origin-ui-core/webpack/base.config.js
```diff
@@ -1,8 +1,11 @@
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
+const Dotenv = require('dotenv-webpack');
```
kosecki123: This whole env thing is not necessary, we approached env settings in a different way. We are using this script https://github.com/energywebfoundation/origin/blob/f6d77f327a7676c3e742cc8a022e5c085cf66e39/packages/origin-ui/env.sh to generate `env-config.js` file which then is read by the UI.

The main reason was to have `.env` variable available on the runtime, including edits. Otherwise full UI repack is required to have your new `.env` variables available.

packages/origin-ui-core/src/features/general/sagas.ts
```diff
@@ -388,13 +387,12 @@ function* fetchDataAfterConfigurationChange(
...
         null
     );
-
     const available = yield all(
         offChainCertificates.balances.available.map((asset) =>
             call(findEnhancedCertificate, asset, initializedCertificates)
         )
     );
-
+    console.groupEnd();
```
kosecki123: ☝️

.env.example
josipbagaric: You mean `DEFAULT_ENERGY_IN_BASE_UNIT`?

I assume "base unit" is Wh, and "display unit" is MWh.

Depends on how this is being used. I see that it's set as a default for an energy input field.
I would set it to 1 000 000 (1 MWh), as we only use and show MWhs in the UI for now.

kosecki123 (approved):

josipbagaric (approved):

PR #1142, opened 2020-06-09
Bug/fix withdrawal test

JGiter:
The test `should be able to withdraw after confirming deposit` in `exchange` package not passing due to the fact that a certificate transaction was added to the transfer repository while transactions on the blockchain was not performed as well as the creation of the certificate itself. To fix the error a certificate was issued and sent to the exchange using issuer.requestCertificationFor and issuer.approveCerificationRequest

### DIFF
--- a/packages/exchange/test/api-orders.e2e-spec.ts
+++ b/packages/exchange/test/api-orders.e2e-spec.ts
@@ -1,9 +1,11 @@
 import { INestApplication } from '@nestjs/common';
 import { expect } from 'chai';
-import { ethers } from 'ethers';
+import { ethers, Wallet, Contract, Event as BlockchainEvent } from 'ethers';
 import request from 'supertest';
 import { OrderStatus } from '@energyweb/exchange-core';
 
+import { ConfigService } from '@nestjs/config';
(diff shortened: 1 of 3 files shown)


packages/exchange/test/api-orders.e2e-spec.ts
```diff
@@ -149,14 +160,49 @@ describe('account ask order send', () => {
     });
 
     it('should be able to withdraw after confirming deposit', async () => {
-        await confirmDeposit();
-
+        const { generationFrom, generationTo, deviceId } = dummyAsset;
```
kosecki123: This is good, but there is already a function in `utils.ts` that does that for you:  https://github.com/energywebfoundation/origin/blob/bb309cae2ed8af868568f82b98f1ae7d739d005f/packages/exchange/test/utils.ts#L8 so please use it if possible.

kosecki123 (approved):

packages/exchange/test/api-orders.e2e-spec.ts
JGiter: Thanks. Its simplifies things a lot

PR #1149, opened 2020-06-10
fix(origin-backend): syntax not null and default

josipbagaric:
Fixes a syntax error when running migrations.


JGiter (changes requested):
`Error: invalid response - 0` while deploying contracts

josipbagaric:
> `Error: invalid response - 0` while deploying contracts

I don't get this error. (running `yarn run:origin`)

kosecki123 (approved):

PR #1152, opened 2020-06-10
chore: move Certification Request validation to the backend

### LINKED ISSUE #1013: Certification request process takes too long to create.
Currently, whenever we need to create a new certification request. `CertificationRequest.getAll()` will be called every time and trying to sync all of the certificate requests.
This process might take a large amount of processing time once we have a lot of certificate requests on our system.

https://github.com/energywebfoundation/origin/blob/b3682c244ce29722a594c1eb13d8b636a447499d/packages/issuer/src/blockchain-facade/CertificationRequest.ts#L196

josipbagaric:
- [x] Create a generation time validation method on the backend (to verify that a certificate for this device and timeframe doesn't already exist)
- [x] e2e tests
- [x] Replace the old frontend validation with the new backend one

Fixes #1013

### DIFF
--- a/packages/issuer/src/blockchain-facade/CertificationRequest.ts
+++ b/packages/issuer/src/blockchain-facade/CertificationRequest.ts
@@ -1,5 +1,3 @@
-import * as Moment from 'moment';
-import { extendMoment } from 'moment-range';
 import { BigNumber, Interface } from 'ethers/utils';
 import { Event as BlockchainEvent } from 'ethers';
 import polly from 'polly-js';
@@ -63,6 +61,8 @@ export class CertificationRequest extends PreciseProofEntity implements ICertifi
             );
         }
 
+        const { certificateClient } = configuration.offChainDataSource;
+
         const request = new CertificationRequest(null, configuration, isVolumePrivate);
 
         const { issuer } = configuration.blockchainProperties as Configuration.BlockchainProperties<
@@ -71,11 +71,15 @@ export class CertificationRequest extends PreciseProofEntity implements ICertifi
         >;
         const issuerWithSigner = issuer.connect(configuration.blockchainProperties.activeUser);
 
-        await this.validateGenerationPeriod(fromTime, toTime, deviceId, configuration);
+        await certificateClient.validateGenerationPeriod({ fromTime, toTime, deviceId });
 
-        const success = await configuration.offChainDataSource.certificateClient.queueCertificationRequestData(
-            { deviceId, fromTime, toTime, energy, files }
-        );
+        const success = await certificateClient.queueCertificationRequestData({
+            deviceId,
+            fromTime,
+            toTime,
+            energy,
+            files
+        });
 
         if (!success) {
             throw new Error('Unable to create CertificationRequest');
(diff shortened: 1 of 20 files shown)


JGiter (approved):

packages/issuer/src/blockchain-facade/CertificationRequest.ts
```diff
@@ -71,11 +71,15 @@ export class CertificationRequest extends PreciseProofEntity implements ICertifi
         >;
         const issuerWithSigner = issuer.connect(configuration.blockchainProperties.activeUser);
 
-        await this.validateGenerationPeriod(fromTime, toTime, deviceId, configuration);
+        await certificateClient.validateGenerationPeriod({ fromTime, toTime, deviceId });
 
-        const success = await configuration.offChainDataSource.certificateClient.queueCertificationRequestData(
-            { deviceId, fromTime, toTime, energy, files }
-        );
+        const success = await certificateClient.queueCertificationRequestData({
```
kosecki123, line 76: Why this actually needs 2 requests? IMHO this should be part of requesting call and request endpoint would need to test that before storing request.

I'm fine with having this as a external call, if in the future we want to use it to make preflight query for UX reasons.

kosecki123 (changes requested):
Missing validation on https://github.com/energywebfoundation/origin/blob/fb1f4cd5c857d529b5b1652dde8b2f02f6accd5b/packages/origin-backend/src/pods/certificate/certification-request.service.ts#L64

packages/issuer/src/blockchain-facade/CertificationRequest.ts
josipbagaric: Good point, tbh I totally overlooked this, this was a linter change.

I'm fine with having it both ways. The only potential issue I see in merging those two requests is that the user might have to wait for the file upload before validation would even start.

Would need to look deeper to see if that's avoidable.

josipbagaric:
> Missing validation on
> 
> https://github.com/energywebfoundation/origin/blob/fb1f4cd5c857d529b5b1652dde8b2f02f6accd5b/packages/origin-backend/src/pods/certificate/certification-request.service.ts#L64

Not sure if this is intentional, it seems to be pointing to code that is not related to this PR.

kosecki123:
> > Missing validation on
> > https://github.com/energywebfoundation/origin/blob/fb1f4cd5c857d529b5b1652dde8b2f02f6accd5b/packages/origin-backend/src/pods/certificate/certification-request.service.ts#L64
> 
> Not sure if this is intentional, it seems to be pointing to code that is not related to this PR.

I wanted to point out the place where `validateGenerationPeriod` function should be added.

packages/issuer/src/blockchain-facade/CertificationRequest.ts
kosecki123: Ah ok, due to file uploads I'm fine with having this additional request. Still same `validateGenerationPeriod` should be executed when making actual request.

kosecki123 (approved):

PR #1153, opened 2020-06-11
feat(origin-ui-core): Add deposit action

JGiter:


packages/localization/src/languages/pl.json
```diff
@@ -78,7 +78,12 @@
                 "returnToInbox": "Zwróć do skrzynki",
                 "buy": "Kup",
                 "request": "Żądaj",
-                "withdraw": "Wycofać"
+                "withdraw": "Wycofać",
```
kosecki123: Please use “Wypłać”

packages/localization/src/languages/pl.json
```diff
@@ -78,7 +78,12 @@
...
+                "deposit": "Kaucja"
```
kosecki123: Please use “Wpłać”

packages/origin-ui-core/src/features/certificates/sagas.ts
```diff
@@ -288,6 +288,63 @@ function* requestPublishForSaleSaga(): SagaIterator {
...
 }
 
+function* requestDepositSaga(): SagaIterator {
+    while (true) {
+        const action: IRequestPublishForSaleAction = yield take(
+            CertificatesActions.requestDepositCertificate
+        );
+        const exchangeClient: IExchangeClient = yield select(getExchangeClient);
+
+        if (!exchangeClient.getAllTransfers) {
```
kosecki123: Wondering about this check. Why we need that?

packages/origin-ui-core/src/features/certificates/sagas.ts
```diff
@@ -288,6 +288,63 @@ function* requestPublishForSaleSaga(): SagaIterator {
...
+
+        yield put(setLoading(true));
+
+        try {
+            const account: ExchangeAccount = yield call([
+                exchangeClient,
+                exchangeClient.getAccount
+            ]);
+
+            if (source === CertificateSource.Blockchain) {
```
kosecki123: I guess we don’t need that check here, or if we need then move it up as a guard closure:

If (source !== ...)...

kosecki123 (approved):
LGTM

packages/origin-ui-core/src/features/certificates/sagas.ts
JGiter: It left after my experiments with certificates. Deleted

packages/origin-ui-core/src/features/certificates/sagas.ts
JGiter: Fixed

PR #1154, opened 2020-06-12
Feature/Bundle market

JGiter:
- add `BundleTable` component
- bundle list is paginated with 25 bundles per page
- price and energy shares are formatted
- add `viewBundleDetails` action
- add react hooks linting
- sort bundles by price and total energy
- view bundle details on bundle click


packages/localization/src/languages/pl.json
```diff
@@ -19,6 +19,9 @@
                 "registerDevice": "Zarejestruj urządzenie",
                 "registerDeviceGroup": "Zarejestruj grupę urządzeń",
                 "pending": "Oczekujące"
+            },
+            "certificates": {
+                "bundles": "Wwiązki"
```
kosecki123: Use "Pakiety"

packages/origin-ui-core/src/components/bundles/BundleTable.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { getConfiguration, getProducingDevices } from '../..';
+import { Bundle } from '../../utils/exchange';
+import { getUserOffchain } from '../../features/users/selectors';
+import BN from 'bn.js';
+import { Visibility } from '@material-ui/icons';
+import { BundleDetails } from './BundleDetails';
+import { getCurrencies } from '../../features';
+import { getBundles } from '../../features/bundles/selectors';
+
+const BUNDLES_PER_PAGE = 25;
```
josipbagaric, line 20: We already have DEFAULT_PAGE_SIZE defined in https://github.com/energywebfoundation/origin/blob/a72cc1a51430c78a61aedc9aa23ea415aeae63a7/packages/origin-ui-core/src/components/Table/PaginatedLoaderHooks.tsx#L11

Can't we just omit this and it will default to 25?

JGiter: DEFAULT_PAGE_SIZE in Table component defined for all Table in the application and can be changed in the future while DEFAULT_PAGE_SIZE in `BundlesTable` should stay 25

josipbagaric: Is there a reason that it should always be 25 and not changeable?

JGiter: Its required in task description - https://energyweb.atlassian.net/browse/PTTDEV-693

josipbagaric (approved):

PR #1155, opened 2020-06-12
fix(origin-backend): fix updating device

josipbagaric:
Fixes an issue where the device wouldn't save changes if there is a certification request present.

### DIFF
--- a/packages/origin-backend/src/pods/device/device.service.ts
+++ b/packages/origin-backend/src/pods/device/device.service.ts
@@ -240,8 +240,10 @@ export class DeviceService {
             throw new NotFoundException(StorageErrors.NON_EXISTENT);
         }
 
+        const { defaultAskPrice, automaticPostForSale } = update;
+
         try {
-            await this.repository.update(id, update);
+            await this.repository.update(id, { defaultAskPrice, automaticPostForSale });
 
             return {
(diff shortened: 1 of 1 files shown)


packages/origin-backend/src/pods/device/device.service.ts
```diff
@@ -240,20 +240,15 @@ export class DeviceService {
...
         }
 
-        device.automaticPostForSale = update.automaticPostForSale;
-        if (update.automaticPostForSale) {
-            device.defaultAskPrice = update.defaultAskPrice;
-        }
-
         try {
-            await this.repository.save((device as unknown) as DeepPartial<Device>);
+            await this.repository.update(id, update);
```
kosecki123: Please change this to an explicit 

{
automaticPostForSale: update.automaticPostForSale
...
} 

accepting input directly is potential security issue, as update can contain more data, malicious one like "owner".

josipbagaric: Available properties are defined in the type: DeviceSettingsUpdateData

josipbagaric: But yeah, it might still happen.

kosecki123 (approved):

PR #1156, opened 2020-06-12
chore: only pull device.meterStats when explicitly requested

josipbagaric:
Every time we would fetch the device from `origin-backend`, the device would be appended with `device.meterStats` regardless on whether we need meterStats or not.

This causes performance issues when `meterStats` are fetched from external APIs.

This PR makes fetching `meterStats` on-demand instead of default.

### DIFF
--- a/packages/device-registry/src/blockchain-facade/ProducingDevice.ts
+++ b/packages/device-registry/src/blockchain-facade/ProducingDevice.ts
@@ -146,15 +146,13 @@ export class Entity implements IDevice {
     }
 }
 
-export const getAllDevices = async (withMeterStats: boolean = false, configuration: Configuration.Entity): Promise<Entity[]> => {
+export const getAllDevices = async (
+    configuration: Configuration.Entity,
+    withMeterStats = false
+): Promise<Entity[]> => {
     const allDevices = await configuration.offChainDataSource.deviceClient.getAll(withMeterStats);
 
     return allDevices.map((device) => new Entity(device.id, configuration, device));
-};
-
-export const getDeviceListLength = async (configuration: Configuration.Entity) => {
-    const allDevices = await configuration.offChainDataSource.deviceClient.getAll(false);
-    return allDevices.length;
 };
 
 export const createDevice = async (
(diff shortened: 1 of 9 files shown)


packages/device-registry/src/blockchain-facade/ProducingDevice.ts
```diff
@@ -146,14 +146,14 @@ export class Entity implements IDevice {
...
+export const getAllDevices = async (withMeterStats: boolean = false, configuration: Configuration.Entity): Promise<Entity[]> => {
+    const allDevices = await configuration.offChainDataSource.deviceClient.getAll(withMeterStats);
 
     return allDevices.map((device) => new Entity(device.id, configuration, device));
 };
 
 export const getDeviceListLength = async (configuration: Configuration.Entity) => {
-    const allDevices = await configuration.offChainDataSource.deviceClient.getAll();
+    const allDevices = await configuration.offChainDataSource.deviceClient.getAll(false);
     return allDevices.length;
```
kosecki123: I know this is not part of this PR, but construct like that should be replaced/avoided. Maybe we should consider removing that method, it's not used aside tests.

packages/origin-ui-core/src/components/AutoSupplyDeviceTable.tsx
```diff
@@ -60,7 +60,7 @@ export function AutoSupplyDeviceTable() {
                 parseInt(requestedFilters[1]?.selectedValue, 10) || 0
             );
         } else {
-            entities = await deviceClient.getMyDevice();
+            entities = await deviceClient.getMyDevices(true);
```
kosecki123: I think we don't need SM reads here.

josipbagaric: ~They are used in that screen:~

Screenshot 2020-06-15 at 10 12 00

EDIT: Nevermind this, this is a different view. Something is weird here. It seems to me that devices should be fetched from the state and not from the client.

kosecki123 (approved):

PR #1179, opened 2020-06-16
Fixes: Admin Table

josipbagaric:
- Simplifies backend for fetching admin panel
- Use **names in enums instead of strings**
- Initialize `kycStatus` for users in seed.sql
- `ActiveUserGuard` now only allows Active users instead of disallowing Pending users
- Remove unnecessary AdminService
- Show warning that user has to be Active to register a device


kosecki123 (approved):
This is surprisingly a lot of changes.

PR #1196, opened 2020-06-22
Matching Engine price picking strategies

kosecki123:
This PR provides:
- abstraction over price picking strategy in the matching engine
- new price picking strategy `OrderCreationTimePickStrategy` which takes the price of the counter order
- configuration item `EXCHANGE_PRICE_STRATEGY` that allow us to chose which price picking strategy should be used 

### OrderCreationTimePickStrategy  examples
1) Given **bids**

| Time   |      Price      |
|----------|------:|
| 1 | $1600 |
| 2 |  $12 |
| 3 | $1 |

arriving **ask** with price $2000 at time later than 3 results in **trade** based on the **bid** price of $1600

2) Given **asks**

| Time   |      Price      |
|----------|------:|
| 1 | $1 |
| 2 |  $12 |
| 3 | $1600 |

arriving **bid** with price $2000 at time later than 3 results in **trade** based on the **ask** price of $1

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -45,6 +45,8 @@ EXCHANGE_WALLET_PRIV=0xd9066ff9f753a1898709b568119055660a77d9aae4d7a4ad677b8fb3d
 EXCHANGE_MATCHING_INTERVAL=1000
 # Issuer ID type
 ISSUER_ID=Issuer ID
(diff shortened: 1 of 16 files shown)


josipbagaric (approved):
👍

PR #1201, opened 2020-06-24
bug/Empty Bundle details

JGiter:
Empty bundle details when bundle volume at least 1000 MWh

When user clicks on bundle to see its details no Bundle Card is rendered and the price range selector consists only from the NaN value. 
The bug was caused by using formatted price (with comma separated decimal groups) to calculate total bundle price


kosecki123:
What do you mean by this PR description?

josipbagaric (approved):

kosecki123 (approved):

PR #1204, opened 2020-06-26
bug/Use bundle splits in BundleDetails

JGiter:


packages/origin-ui-core/src/utils/device.ts
```diff
@@ -216,24 +224,24 @@ export function useDevicePermissions() {
 export const deviceById = (
     id: string,
     environment: IEnvironment,
-    devices: ProducingDevice.Entity[]
+    device: ProducingDevice.Entity[]
 ): ProducingDevice.Entity => {
-    return devices.find((d) => {
+    return device.find((d) => {
```
josipbagaric: Not sure why this was renamed? Array === plural

josipbagaric (approved):

packages/origin-ui-core/src/utils/device.ts
JGiter: Fixed

kosecki123 (approved):

PR #1206, opened 2020-06-26
fix(origin-backend): fix device permissioning - storing reads and upd…

josipbagaric:
Fixes the following security issues:
- storing smart meter reads can be done by anyone
- any organization admin can update any device status - only issuer/paltformadmin should be able to do that
- any org admin or device manager can delete any device


kosecki123:
I think this will make solar simulator to not be able to store any reads for devices.

JGiter (approved):

PR #1228, opened 2020-07-01
bug/request account with no account id

JGiter:
Issue: When user which doesn't belong to any organization log in backend return 500 on attempt to get user account

Reason: Application expects that the logged in user has exchange account and this account is bound to user's organization id

Fix: Use userId as its owner id when organization id is not exist


josipbagaric (approved):
I guess we can remove the comment above the code.

PR #1251, opened 2020-07-06
feature/my open position

JGiter:
- Added My Orders tab with My Asks and My Bids tables
- Tables rows are filtered by generation time period, device type and facility name


packages/exchange/src/pods/order/product.dto.ts
```diff
@@ -38,6 +38,9 @@ export class ProductDTO {
     @Validate(GridOperatorValidator)
     public gridOperator?: string[];
 
+    @IsOptional()
+    public externalDeviceId?: { id: string; type: string };
```
josipbagaric: Replace with the proper type `IExternalDeviceId`.

https://github.com/energywebfoundation/origin/blob/52e5fccdeb303133b3db9413fc317e4a1c60ebca/packages/origin-backend-core/src/Device.ts#L10

packages/origin-ui-core/src/components/Modal/AskDetailslModal.tsx
```diff
@@ -0,0 +1,17 @@
...
+interface IOwnProps {
+    ask: Order;
+    close: () => void;
+}
+
+export const AskDetailsModal = (props: IOwnProps) => {
+    const { ask, close } = props;
+    return (
+        <Dialog open={ask !== null} onClose={close}>
+            View ask details modal
```
josipbagaric, line 14: Needs to be localized.

packages/origin-ui-core/src/components/Modal/BidDetalisModal.tsx
```diff
@@ -0,0 +1,17 @@
...
+interface IOwnProps {
+    bid: Order;
+    close: () => void;
+}
+
+export const BidsDetailsModal = (props: IOwnProps) => {
+    const { bid, close } = props;
+    return (
+        <Dialog open={bid !== null} onClose={close}>
+            View ask details modal
```
josipbagaric, line 14: Localize

packages/origin-ui-core/src/components/Modal/RemoveOrderConfirmation.tsx
```diff
@@ -0,0 +1,17 @@
...
+interface IOwnProps {
+    order: Order;
+    close: () => void;
+}
+
+export const RemoveOrderConfirmation = (props: IOwnProps) => {
+    const { order, close } = props;
+    return (
+        <Dialog open={order !== null} onClose={close}>
+            View ask details modal
```
josipbagaric, line 14: Localize

packages/origin-ui-core/src/components/Table/PaginatedLoaderHooks.tsx
```diff
@@ -205,10 +206,13 @@ export function checkRecordPassesFilters(
...
     for (const filter of filters) {
         const filteredPropertyResolvedValue = filter.property(record);
-
         if (typeof filteredPropertyResolvedValue !== 'undefined') {
             switch (filter.input.type) {
                 case CustomFilterInputType.string:
+                    console.log(
+                        '>>> String filter: record to filter:',
+                        filteredPropertyResolvedValue
+                    );
```
josipbagaric: Do we need to keep these console logs?

packages/origin-ui-core/src/components/orders/AsksTable.tsx
```diff
@@ -0,0 +1,198 @@
...
+    const removeAsk = (rowIndex: number) => {
+        const { askId } = rows[rowIndex];
+        const ask = asks.find((o) => o.id === askId);
+        setToRemove(ask);
+    };
+
+    const actions = [
+        {
+            icon: <Visibility />,
+            name: 'View',
```
josipbagaric: Localize

packages/origin-ui-core/src/components/orders/AsksTable.tsx
```diff
@@ -0,0 +1,198 @@
...
+            onClick: (row: string) => viewDetails(parseInt(row, 10))
+        },
+        {
+            icon: <Remove />,
+            name: 'Remove',
```
josipbagaric: Localize

packages/origin-ui-core/src/components/orders/BidsTable.tsx
```diff
@@ -0,0 +1,193 @@
...
+    const removeBid = (rowIndex: number) => {
+        const { bidId } = rows[rowIndex];
+        const bid = bids.find((o) => o.id === bidId);
+        setToRemove(bid);
+    };
+
+    const actions = [
+        {
+            icon: <Visibility />,
+            name: 'View',
```
josipbagaric: Localize?

packages/origin-ui-core/src/components/orders/BidsTable.tsx
```diff
@@ -0,0 +1,193 @@
...
+            onClick: (row: string) => viewDetails(parseInt(row, 10))
+        },
+        {
+            icon: <Remove />,
+            name: 'Remove',
```
josipbagaric: Localize

packages/origin-ui-core/src/styles/variables.scss
```diff
@@ -21,6 +21,8 @@ $createBundleCertificatesGroupBg: #3b3b3b;
 $createBundleCertificatesGroupHeaderBg: #373737;
 $createBundleCertificateForSale: #303030;
 $createBundleSelectedCertificate: #3a2146;
+$filterBg: #3a3a3a;
+$filterItemBg: #2d2d2d;
```
josipbagaric: **Do we really need these many new colors**? Can't we replace most of these colors with a few colors which can be reused?

1. Are `$filterBg` and `$createBundleCertificatesGroupBg` really that different that they have to be defined differently?
2. `$bundleDetailsBg` and `$bundleCardBorderBg` are basically the **same color**, why are they defined twice?
3. `$filterItemBg` and `$createBundleCertificatesBg` are the **same color**.

Can't these just be replaced with one color called `$secondaryColor: #3b3b3b`?

Then we can create different colors such as `$secondaryColorDarker: darken($secondaryColor, 10%)`.

Adding a new color for each new component that we build is stupid. We should have a set number of colors in the whole application, which we then reuse, and that can be easily customized.

I suggest a similar approach to [Boostrap](https://getbootstrap.com/docs/4.0/utilities/colors/#background-color), where we can define a **primary** (e.g. purple) and **secondary** color (e.g. grey), and then all other variations of that color is dynamically generated using `darken()` and `lighten()` variations of the primary and secondary colors.

Does this make sense?

packages/origin-ui-core/src/utils/exchange/types.ts
```diff
@@ -13,6 +13,7 @@ export interface IProductDTO {
     generationFrom?: string;
     generationTo?: string;
     gridOperator?: string[];
+    externalDeviceId?: { id: string; type: string };
```
josipbagaric: Replace with the proper type `IExternalDeviceId`.

https://github.com/energywebfoundation/origin/blob/52e5fccdeb303133b3db9413fc317e4a1c60ebca/packages/origin-backend-core/src/Device.ts#L10

josipbagaric (approved):

josipbagaric (changes requested):
Please check the review for necessary changes.

packages/origin-ui-core/src/components/Modal/AskDetailslModal.tsx
JGiter: These modals (Ask and Bids details and remove) are templates and developed in other PR

packages/origin-ui-core/src/components/Table/PaginatedLoaderHooks.tsx
JGiter: sorry. I am just keep to forget to remove consols. May be we should add eslint rule which forbids them

packages/origin-ui-core/src/styles/variables.scss
JGiter: Agree. I'll try to implement this approach

packages/origin-ui-core/src/components/Modal/BidDetalisModal.tsx
JGiter: Its temporary templates

packages/origin-ui-core/src/components/Modal/RemoveOrderConfirmation.tsx
JGiter: Temporary templates

packages/origin-ui-core/src/components/Table/PaginatedLoaderHooks.tsx
josipbagaric: We thought about it, but we do have intentional console.logs in the codebase, so it wouldn't work.

josipbagaric (approved):

PR #1260, opened 2020-07-10
bug(origin-ui-core): My orders are not displayed

JGiter:
Fixed bug: After publishing certificate it doesn't appears in My Orders tab

Reason: As order id a user id was used

### DIFF
--- a/packages/origin-ui-core/src/components/orders/MyOrders.tsx
+++ b/packages/origin-ui-core/src/components/orders/MyOrders.tsx
@@ -1,18 +1,16 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
-import { OrderSide } from '@energyweb/exchange-core';
+import { OrderSide, OrderStatus } from '@energyweb/exchange-core';
 
 import { Order } from '../../utils/exchange';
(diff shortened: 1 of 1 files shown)


packages/origin-ui-core/src/components/orders/MyOrders.tsx
```diff
@@ -11,7 +11,9 @@ import { Box, useTheme } from '@material-ui/core';
 
 export const MyOrders = () => {
     const user = useSelector(getUserOffchain);
-    const orders: Order[] = useSelector(getOrders).filter((o) => o.userId === user.id.toString());
+    const orders: Order[] = useSelector(getOrders).filter(
+        (o) => o.userId === user.organization.id.toString()
```
kosecki123: Are you sure we need this .filter ? When requesting using GET /orders you will receive **your** orders anyway.

JGiter: You right. I'll remove it

kosecki123 (approved):

josipbagaric (approved):

PR #1290, opened 2020-07-15
bug(origin-ui-core): fetch unnecessary splits

JGiter:
Fixed performance issue: On loading application all bundle splits are loaded without need. 
Current behavior: splits loaded on bundle details

### DIFF
--- a/packages/origin-ui-core/src/components/Modal/BundleBougth.tsx
+++ b/packages/origin-ui-core/src/components/Modal/BundleBougth.tsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import {
+    DialogTitle,
+    DialogActions,
+    Button,
+    Dialog,
+    DialogContent,
+    makeStyles,
+    Theme,
+    Box,
+    Grid
+} from '@material-ui/core';
+import { RemoveCircle } from '@material-ui/icons';
+import { useTranslation } from '../..';
+
+interface IOwnProps {
+    open: boolean;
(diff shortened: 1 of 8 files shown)


packages/origin-ui-core/src/components/Modal/BundleBougth.tsx
```diff
@@ -0,0 +1,53 @@
...
+                        />
+                    </Grid>
+                    <Grid item>
+                        <Box>{t('bundle.info.bundleBought')}</Box>
+                    </Grid>
+                </Grid>{' '}
+            </DialogContent>
+            <DialogActions>
+                <Button onClick={() => setOpen(false)} color="primary">
+                    Ok
```
josipbagaric: I guess we should also localize "OK". It might not be available in some languages (Japanese, Chinese, Thai, etc.).

packages/origin-ui-core/src/components/bundles/BundlesTable.tsx
```diff
@@ -158,6 +173,7 @@ export const BundlesTable = (props: IOwnProps) => {
                     </Fab>
                 </Tooltip>
             </Link>
+            <BundleBougth open={showBundleBoughtModal} setOpen={setShowBundleBougthModal} />
```
josipbagaric: Typo `BundleBougth`. Use `BundleBought` instead.

kosecki123 (approved):

PR #1291, opened 2020-07-16
bug/forbidden acces to organization

JGiter:
Fixed bug:
- User without admin rights trying to fetch list of organizations which is needed to accept invitation;
Fix: Use organization information from invitation;

Performance improvement: fetch organization list when user navigates on Organization tab;

### DIFF
--- a/packages/origin-ui-core/src/components/Organization/Organization.tsx
+++ b/packages/origin-ui-core/src/components/Organization/Organization.tsx
@@ -1,19 +1,17 @@
-import React, { useState } from 'react';
-import { useSelector, useDispatch } from 'react-redux';
+import React from 'react';
+import { useSelector } from 'react-redux';
 import { NavLink, Route, Redirect } from 'react-router-dom';
-import { Role, isRole, IOrganizationWithRelationsIds } from '@energyweb/origin-backend-core';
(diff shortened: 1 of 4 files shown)


packages/origin-ui-core/src/components/Organization/Organization.tsx
```diff
@@ -20,24 +22,48 @@ export const roleNames = {
...
+        setFetching(false);
+    };
+
+    if (
+        user &&
+        (!storedOrganizations || storedOrganizations.length === 0) &&
+        !fetching &&
+        isRole(user, Role.Admin, Role.SupportAgent)
+    ) {
+        fetchOrganizations();
```
josipbagaric: Does this need to be fetched in this Organization component? I think this is only needed in the child `OrganizationTable` component.

kosecki123 (approved):

packages/origin-ui-core/src/components/Organization/Organization.tsx
JGiter: You are right. Even in `OrganizationTable` its not needed as it accessible based on role

josipbagaric (approved):

PR #1300, opened 2020-07-17
bug/read certificates based on user profile

JGiter:
Certificates are read based on address in user profile

### DIFF
--- a/packages/issuer/src/blockchain-facade/CertificateUtils.ts
+++ b/packages/issuer/src/blockchain-facade/CertificateUtils.ts
@@ -7,6 +7,7 @@ import { Certificate, IClaimData } from './Certificate';
 import { Registry } from '../ethers/Registry';
 import { Issuer } from '../ethers/Issuer';
 import { getEventsFromContract } from '../utils/events';
+import { IUserWithRelations } from '../../../origin-backend-core/src';
 
 export interface IBlockchainEvent {
     name: string;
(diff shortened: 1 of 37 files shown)


packages/issuer/src/blockchain-facade/CertificateUtils.ts
```diff
@@ -7,6 +7,7 @@ import { Certificate, IClaimData } from './Certificate';
 import { Registry } from '../ethers/Registry';
 import { Issuer } from '../ethers/Issuer';
 import { getEventsFromContract } from '../utils/events';
+import { IUserWithRelations } from '../../../origin-backend-core/src';
```
kosecki123: This should be imported from '@energyweb/origin-backend-core'

packages/issuer/src/blockchain-facade/CertificateUtils.ts
```diff
@@ -171,13 +172,13 @@ export async function getAllCertificates(
 }
 
 export async function getAllOwnedCertificates(
-    configuration: Configuration.Entity
+    configuration: Configuration.Entity,
+    user: IUserWithRelations
```
kosecki123: What's the reason of adding extra `user` paramater, where configuration has already all necessary fields? Seem to me that this PR should just change configuration.activeUser to be set from config instead of MetaMask.

kosecki123 (changes requested):
As in comments.

kosecki123 (approved):

PR #1314, opened 2020-07-20
chore: migrate ethers.js to version 5.x

josipbagaric:
Update Origin to use the latest version of the Ethers.js library.

Closes https://github.com/energywebfoundation/origin/pull/1169.

### DIFF
--- a/packages/exchange/test/utils.ts
+++ b/packages/exchange/test/utils.ts
@@ -22,11 +22,14 @@ export const issueToken = async (
         false
     )) as ContractTransaction).wait();
 
-    const {
-        values: { _id: requestId }
-    } = registryInterface.parseLog(requestReceipt.logs[0]);
+    const [log] = requestReceipt.logs;
 
-    const validityData = registryInterface.functions.isRequestValid.encode([requestId.toString()]);
+    const { name } = registryInterface.parseLog(log);
+    const { _id: requestId } = registryInterface.decodeEventLog(name, log.data, log.topics);
+
+    const validityData = registryInterface.encodeFunctionData('isRequestValid', [
+        requestId.toString()
+    ]);
 
     const approvalReceipt = await ((await issuer.approveCertificationRequest(
         requestId,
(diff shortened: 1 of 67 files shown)


packages/exchange/test/utils.ts
```diff
@@ -48,13 +51,7 @@ export const depositToken = async (
...
     const registryWithUserAsSigner = registry.connect(sender);
 
-    await registryWithUserAsSigner.functions.safeTransferFrom(
-        sender.address,
-        to,
-        id,
-        amount,
-        '0x0'
-    );
+    await registryWithUserAsSigner.safeTransferFrom(sender.address, to, id, amount, '0x00');
```
kosecki123, line 54: Is there a literal for '0x00' in the ether lib? Seems for some reason we need to use '0x00' instead of '0x0'

packages/issuer/contracts/Issuer.sol
```diff
@@ -60,9 +60,9 @@ contract Issuer is Initializable, Ownable {
 		bytes32 hash;
 	}
 
-    function initialize(int _certificateTopic, address _registry, address _owner) public initializer {
-        require(_registry != address(0), "initialize: Cannot use address 0x0 as registry address.");
-        require(_owner != address(0), "initialize: Cannot use address 0x0 as the owner.");
+    function init(int _certificateTopic, address _registry, address _owner) public initializer {
```
kosecki123: Why is the rename applied? This requires us to redeploy etc.

packages/exchange/test/utils.ts
josipbagaric: Ethers no longer accepted `0x0` because it's an odd hex number. Not sure why that change happened.

kosecki123 (approved):

packages/issuer/contracts/Issuer.sol
josipbagaric: For some reason the function with name `initialize` was not available in the typechain autogenerated facade. I renamed it and it worked (dunno why, figured it was a naming conflict with ethers).

But a good point to re-review this. I might try something like this instead: https://github.com/ethers-io/ethers.js/issues/407#issuecomment-458329708.

PR #1315, opened 2020-07-21
bug/login doesnt reset user info

JGiter:
Fixed issue:
- [x] After changed logged in user list of his owned bundles doesn't updates

### DIFF
--- a/packages/origin-ui-core/src/features/users/sagas.ts
+++ b/packages/origin-ui-core/src/features/users/sagas.ts
@@ -15,8 +15,8 @@ import {
     IOrganizationWithRelationsIds
 } from '@energyweb/origin-backend-core';
 import { GeneralActions, ISetOffChainDataSourceAction } from '../general/actions';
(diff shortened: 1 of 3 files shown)


packages/origin-ui-core/src/features/users/sagas.ts
```diff
@@ -103,6 +104,7 @@ function* fetchOffchainUserDetails(): SagaIterator {
                     organization
                 })
             );
+            yield put(reloadCertificates());
```
kosecki123, line 109: Is that the correct approach? What is the root cause of having wrong data after logout/login sequence?

JGiter: The reason is that bundles are not refetched after user login. `reloadCertificates` action triggers this reloading along with reloading of the certificates and orders as it would be better to synchronize their state

kosecki123 (approved):

PR #1316, opened 2020-07-21
chore: add support for a backup web3 node

josipbagaric:
Make use of `ethers.providers.FallbackProvider` to create a fallback node in case the primary RPC node goes down.


.env.example
```diff
@@ -19,8 +19,10 @@ JWT_SECRET=thishastobesecret13213
 JWT_EXPIRY_TIME=7 days
 
 # ------- OPTIONAL ------- #
+WEB3_BACKUP=http://localhost:8545
```
kosecki123: Since ethers fallback provider accepts an array of providers, would be to allow user to specify N nodes formatted as

`WEB3=node1url;node2url;...nodeNurl`

Then use the order of this config to assign the prioroties

.env.example
```diff
@@ -2,6 +2,7 @@
 MODE=dev
 
 # URLs
+# You can add multiple WEB3 nodes, separated by a semilcolon ';'. e.g. WEB3=http://localhost:8545;http://localhost:8550
 WEB3=http://localhost:8545
```
kosecki123, line 6: I guess this should be remove as it duplicates line above.

kosecki123 (approved):

.env.example
josipbagaric: The line above is just a comment.
I would leave this line so that people can c/p this .env.example, rename to .env and get going.

kosecki123: Ok, GH wrapped the line after e.g. so it looked for me like

```
WEB3=http://localhost:8545;http://localhost:8550
WEB3=http://localhost:8545
```

PR #1317, opened 2020-07-22
bug(origin-ui-core): cant verify blockchain account of inactive user 

JGiter:
Fixed issue:
- Only active user can verify his blockchain account

Current behavior: Only deleted user is not allowed to verify account

### DIFF
--- a/packages/origin-backend-utils/src/guard/NotDeletedUserGuard.ts
+++ b/packages/origin-backend-utils/src/guard/NotDeletedUserGuard.ts
@@ -1,22 +1,20 @@
 import { Injectable, CanActivate, ExecutionContext, HttpException } from '@nestjs/common';
 import { Reflector } from '@nestjs/core';
-import { IUserWithRelationsIds, UserStatus } from '@energyweb/origin-backend-core';
+import { UserStatus } from '@energyweb/origin-backend-core';
 
 @Injectable()
 export class NotDeletedUserGuard implements CanActivate {
     constructor(private reflector: Reflector) {}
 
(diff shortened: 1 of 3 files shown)


packages/origin-backend-utils/src/guard/NotDeletedUserGuard.ts
```diff
@@ -0,0 +1,25 @@
...
+import { IUserWithRelationsIds, UserStatus } from '@energyweb/origin-backend-core';
+
+@Injectable()
+export class NotDeletedUserGuard implements CanActivate {
+    constructor(private reflector: Reflector) {}
+
+    canActivate(context: ExecutionContext): boolean {
+        const request = context.switchToHttp().getRequest();
+        const user = request.user as IUserWithRelationsIds;
+        const _user = user as IUserWithRelationsIds;
```
kosecki123: This should be 1 line.

packages/origin-backend-utils/src/guard/NotDeletedUserGuard.ts
```diff
@@ -0,0 +1,25 @@
...
+
+        if (_user.status === UserStatus.Deleted) {
+            throw new HttpException(
+                `Only not deleted users can perform this action. Your status is ${
+                    UserStatus[_user.status]
+                }`,
+                412
```
kosecki123: This should be 403 http error code, so you can use ForbiddenExpection or ForbiddenHttpException

kosecki123 (approved):

PR #1322, opened 2020-07-24
feature/post specific volume of bundle item

JGiter:
Added feature: When creating bundle user now can specify volume of each certificate included in bundle instead of including whole cerfificate volume
image

### DIFF
--- a/packages/origin-ui-core/src/components/OriginConfigurationContext.tsx
+++ b/packages/origin-ui-core/src/components/OriginConfigurationContext.tsx
@@ -26,7 +26,7 @@ export interface IOriginStyleConfig {
 
 export const createMaterialThemeForOrigin = (
     styleConfig: IOriginStyleConfig,
-    language: ORIGIN_LANGUAGE,
+    language: ORIGIN_LANGUAGE
 ): Theme => {
     const materialLocale =
         {
@@ -60,11 +60,11 @@ export const createMaterialThemeForOrigin = (
                         }
                     }
                 },
-                MuiFormControl: {
-                    root: {
-                        backgroundColor: styleConfig.BACKGROUND_COLOR_DARKER
-                    }
-                },
+                // MuiFormControl: {
+                //     root: {
+                //         backgroundColor: styleConfig.BACKGROUND_COLOR_DARKER
+                //     }
+                // },
                 MuiFormLabel: {
                     root: { fontSize: variables.fontSizeMd }
                 },
--- a/packages/origin-ui-core/src/components/OriginConfigurationContext.tsx
+++ b/packages/origin-ui-core/src/components/OriginConfigurationContext.tsx
@@ -24,7 +24,7 @@ export interface IOriginStyleConfig {
     WHITE: string;
 }
 
-const DEFAULT_COLOR='#894ec5';
+const DEFAULT_COLOR = '#894ec5';
 
 export const createMaterialThemeForOrigin = (
     styleConfig: IOriginStyleConfig,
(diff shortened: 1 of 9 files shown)


packages/origin-ui-core/src/components/OriginConfigurationContext.tsx
```diff
@@ -59,6 +60,11 @@ export const createMaterialThemeForOrigin = (
                         }
                     }
                 },
+                // MuiFormControl: {
+                //     root: {
+                //         backgroundColor: styleConfig.BACKGROUND_COLOR_DARKER
+                //     }
+                // },
```
josipbagaric: Commented code.

packages/origin-ui-core/src/components/OriginConfigurationContext.tsx
```diff
@@ -201,6 +207,7 @@ export function createStyleConfigFromSCSSVariables(scssVariables: any): IOriginS
         PRIMARY_COLOR_DARKER: scssVariables.primaryColorDarker ?? '#894ec5',
         TEXT_COLOR_DEFAULT: scssVariables.textColorDefault ?? '#894ec5',
         BACKGROUND_COLOR_DARK: scssVariables.backgroundColorDark ?? '#894ec5',
+        BACKGROUND_COLOR_DARKER: scssVariables.backgroundColorDarker ?? '#894ec5',
```
josipbagaric: Why do we keep copying #894ec5? Let's define it once above as a default color and then reference it as a variable.

packages/origin-ui-core/src/components/bundles/BundleItemEdit.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    const { province, deviceType, facilityName } = deviceById(deviceId, environment, devices);
+    const type = deviceType.split(';')[0].toLowerCase() as EnergyTypes;
+
+    const { Yup } = useValidation();
+    const validationSchema = Yup.object().shape({
+        volumeToBundle: Yup.number()
+            .required()
+            .min(0)
+            .max(Number(EnergyFormatter.format(publicVolume)))
```
josipbagaric: Cleaner way of getting the number is `.max(publicVolume.toNumber())` (instead of coverting the number to string and then casting to Number).

packages/origin-ui-core/src/components/bundles/BundleItemEdit.tsx
```diff
@@ -0,0 +1,231 @@
...
+                                                                <IconButton
+                                                                    onClick={() =>
+                                                                        setFieldValue(
+                                                                            'volumeToBundle',
+                                                                            0
+                                                                        )
+                                                                    }
+                                                                >
+                                                                    <HighlightOff
+                                                                        style={{ fill: '#878787' }}
```
josipbagaric: Do we need to have this color defined in the component?

packages/origin-ui-core/src/components/bundles/BundleItemEdit.tsx
JGiter: But we still need `EnergyFormatter.format` which converts Wh to MWh and returns string

packages/origin-ui-core/src/components/bundles/BundleItemEdit.tsx
JGiter: This is the color of the reset button. As I understand it shouldn't change
image

Moved it to BundleItemEdit.scss

packages/origin-ui-core/src/components/bundles/BundleItemEdit.tsx
josipbagaric: For converting between MWh and Wh please use either `EnergyFormatter.getValueInDisplayUnit()` or `EnergyFormatter.getBaseValueFromValueInDisplayUnit()`.

`EnergyFormatter.format()` should only be used to print, not to convert.

josipbagaric (approved):

PR #1323, opened 2020-07-24
chore: export individual NestJS modules

josipbagaric:


kosecki123:
Those modules relies on a global pipes and filters defined in the app.module. We should also export them and enable or document.

kosecki123 (approved):

PR #1338, opened 2020-07-28
bug/(origin-ui-core): Can't login after accepting organization invitation

JGiter:
Fixed: User with `Member` role can't login after accepting organization invitation


packages/origin-ui-core/src/features/users/sagas.ts
```diff
@@ -91,7 +92,10 @@ function* fetchOffchainUserDetails(): SagaIterator {
 
             let organization: IOrganizationWithRelationsIds = null;
 
-            if (typeof userProfile.organization !== 'undefined') {
+            if (
+                typeof userProfile.organization !== 'undefined' &&
+                [Role.Admin || Role.SupportAgent].includes(userProfile.organization)
```
josipbagaric: @JGiter could you explain this line? I'm trying to understand what Roles have to do with the organization ID.

JGiter: In this block the user organization is fetched and if user doesn't have necessary rights the error is thrown

josipbagaric: Why are you comparing Roles to the Organization ID? The organization ID has no connection to Roles.
Roles are checked from `user.rights`, nothing to do with the organization.

You can also use the `isRole()` [helper function](https://github.com/energywebfoundation/origin/blob/1753e04781a422782d58874e8456e81d411cd460/packages/origin-backend-core/src/User.ts#L47).

`isRole(userProfile, Role.Admin, Role.SupportAgent)`.

JGiter: You right. Its typo

josipbagaric (changes requested):
Please use the `isRole()` [helper function](https://github.com/energywebfoundation/origin/blob/1753e04781a422782d58874e8456e81d411cd460/packages/origin-backend-core/src/User.ts#L47) to minimize potential errors and make the code more maintainable and readable.

josipbagaric (approved):

PR #1339, opened 2020-07-28
chore: allow defining features when running exchange and/or origin-ba…

josipbagaric:
Allow running `@energyweb/exchange` and `@energyweb/origin-backend` by defining a set of `OriginFeature`s

### DIFF
--- a/packages/exchange/src/app.module.ts
+++ b/packages/exchange/src/app.module.ts
@@ -1,8 +1,9 @@
-import { Module, ValidationPipe } from '@nestjs/common';
+import { Module, ValidationPipe, DynamicModule } from '@nestjs/common';
 import { ConfigModule } from '@nestjs/config';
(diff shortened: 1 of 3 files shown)


kosecki123:
Do we want to handle this by providing flags? I think we should consider other option, since we are now exporting all modules separately, so the runner module gathers and uses only those which are necessary.

Currently there are many cross-module dependencies, in the long run we should refactor to cqrs/events to decouple them from eachother.

josipbagaric:
We don't have to handle this with flags, these are just to simplify/help with building pilots.
On the pilot side, I can always selectively import specific modules instead of using the flags.

Definitely agree on the refactor into cqrs/events, this might be necessary in future, if this selective module approach won't work.

josipbagaric:
Tested this approach indirectly, it will mot be sufficient. Closing the PR.

Issue #1341, opened 2020-07-29
Make excludeEnd on DemandTimePeriodService modifiable


atchapcyp:
https://github.com/energywebfoundation/origin/blob/e24268d9dbadf0aa070cceb8efdcb54da9e23b5f/packages/exchange/src/pods/demand/demand-time-period.service.ts#L21

Currently, If a user wants to create a monthly demand of 2 months user needs to input Jan, 1 - March, 1. Which is not a good experience when frontend app uses a month-picker component (user need to click on Jan, Feb, and March).

I think it will be better if we allow the service caller decide how to adjust the demand period (frontend friendly).

kosecki123:
Fixed in #1342

PR #1342, opened 2020-07-29
Update excludeEnd for DemandTimePeriodService

atchapcyp:
issue : https://github.com/energywebfoundation/origin/issues/1341


packages/exchange/src/pods/demand/demand-time-period.service.ts
```diff
@@ -18,9 +18,9 @@ export class DemandTimePeriodService {
         const range = moment.range(start, end);
 
         const { diff, step } = this.timeFrameToTimeDiff(createDemand.periodTimeFrame);
-        const validFrom = Array.from(range.by(diff, { step, excludeEnd: true })).map((v) =>
-            v.toDate()
-        );
+        const validFrom = Array.from(
+            range.by(diff, { step, excludeEnd: createDemand.excludeEnd })
```
kosecki123: Since createDemand.excludeEnd would be marked as optional, please change:

`excludeEnd: createDemand.excludeEnd`

to

`excludeEnd: createDemand.excludeEnd ?? true`

packages/exchange/src/pods/demand/create-demand.dto.ts
```diff
@@ -37,4 +37,7 @@ export class CreateDemandDTO {
 
     @IsBoolean()
     public readonly boundToGenerationTime: boolean;
+
+    @IsBoolean()
```
kosecki123, line 42: Please mark this field as Optional() as well, this would allow us to retain the compatibility if previous API deployments. 

Also, the default version used by the system should be `true` as it was previously.

kosecki123 (changes requested):
Please check my comments, otherwise 🚀

kosecki123:
Thanks for the contribution! Please also rebase you commits to follow conventional commits standard, for e.g:

`feat(exchange): add excludeEnd to CreateDemandDTO`

kosecki123:
Fixes #1341

kosecki123 (approved):

PR #1344, opened 2020-07-30
bug/decimal bundle volume

JGiter:
Fixed issue:
Front-end and back-end allows creation of bundle with decimal volume of the items

image

### DIFF
--- a/packages/origin-backend-app/src/main.ts
+++ b/packages/origin-backend-app/src/main.ts
@@ -1,5 +1,5 @@
 import { AppModule as ExchangeModule } from '@energyweb/exchange';
-import { LoggerService } from '@nestjs/common';
+import { LoggerService, ValidationPipe } from '@nestjs/common';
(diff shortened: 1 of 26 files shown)


packages/origin-backend-app/src/main.ts
```diff
@@ -1,5 +1,5 @@
 import { AppModule as ExchangeModule } from '@energyweb/exchange';
-import { LoggerService } from '@nestjs/common';
+import { LoggerService, ValidationPipe } from '@nestjs/common';
```
kosecki123: Are these change necessary? ValidationPipe is already set a provider

https://github.com/energywebfoundation/origin/blob/376b6ba4b3e7487b731e5bcbd9d27d81a4ba4580/packages/exchange/src/app.module.ts#L67

kosecki123 (changes requested):
All good, just please add a test for that case.

kosecki123 (approved):

PR #1346, opened 2020-07-30
bug(origin-backend): User organization is not shown in the invitation table

JGiter:
Fixed issue:
image

### DIFF
--- a/packages/origin-backend-client-mocks/src/OrganizationClientMock.ts
+++ b/packages/origin-backend-client-mocks/src/OrganizationClientMock.ts
@@ -9,7 +9,8 @@ import {
     OrganizationUpdateData,
     OrganizationRole,
     Role,
-    IOrganizationClient
+    IOrganizationClient,
+    IOrganization
 } from '@energyweb/origin-backend-core';
 
 interface ITmpUser {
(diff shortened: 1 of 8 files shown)


packages/exchange/src/pods/account/account.controller.ts
```diff
@@ -19,7 +19,7 @@ export class AccountController {
     // TODO: explicit account creation request
     @UseInterceptors(ClassSerializerInterceptor)
     @Get()
-    @UseGuards(AuthGuard(), ActiveUserGuard)
+    @UseGuards(AuthGuard())
```
kosecki123: Why this has been removed? Does not seems to have anything to do with PR.

packages/exchange/src/pods/bundle/bundle.controller.ts
```diff
@@ -44,7 +44,7 @@ export class BundleController {
     }
 
     @Get()
-    @UseGuards(AuthGuard(), ActiveUserGuard)
+    @UseGuards(AuthGuard())
```
kosecki123: Why this has been removed? Does not seems to have anything to do with PR.

packages/exchange/src/pods/order/order.controller.ts
```diff
@@ -84,7 +84,7 @@ export class OrderController {
     }
 
     @Get()
-    @UseGuards(AuthGuard(), ActiveUserGuard)
+    @UseGuards(AuthGuard())
```
kosecki123: Why this has been removed? Does not seems to have anything to do with PR.

kosecki123 (approved):

PR #1348, opened 2020-07-31
bug/forbid decimal volumes 

JGiter:
Validate that volumes are expressed in energy units in:
 - [x] open orders
 - [x] demands
Because the origin-backend module is dynamic, Nestjs doesn't allow to select it to inject its services into the validator. Therefore validation of certificate request will be done in separate PR


kosecki123:
Rebase you commits again so changes to 'exchange' package and 'origin ui' are not in the same commit.

kosecki123 (approved):

PR #1365, opened 2020-08-06
feature/backend invitation processing

JGiter:
- `IOrganizationInvitation` extended with `sender` property to display invitation sender to the invited user 
- add `Viewed` status to `IOrganizationInvitationStatus` enum to track whether invitation was communicated
- `organization` property type of `IOrganizationInvitation` changed to `IOrganization`
- Changes to `OrganizationInvitation` entity are migrated
- Since invitation updating now is not only about acceptance or rejection `OrganizationInvitationService.acceptOrReject` renamed to `update` and its logic is changed

### DIFF
--- a/packages/origin-backend-core/src/OrganizationInvitation.ts
+++ b/packages/origin-backend-core/src/OrganizationInvitation.ts
@@ -4,7 +4,8 @@ import { Role } from './User';
 export enum OrganizationInvitationStatus {
     Pending,
     Rejected,
-    Accepted
+    Accepted,
+    Viewed
 }
 
 export type OrganizationRole =
@@ -21,6 +22,7 @@ export interface IOrganizationInvitationProperties {
 
 export interface IOrganizationInvitation extends IOrganizationInvitationProperties {
     organization: IOrganization;
+    sender: string;
 }
 
 export type OrganizationInviteCreateData = { email: string; role: OrganizationRole };
--- a/packages/origin-backend-client-mocks/src/OrganizationClientMock.ts
+++ b/packages/origin-backend-client-mocks/src/OrganizationClientMock.ts
@@ -69,15 +69,21 @@ export class OrganizationClientMock implements IOrganizationClient {
         throw new Error('Method not implemented.');
     }
 
-    inviteMocked(email: string, organizationId: number, role: OrganizationRole): ISuccessResponse {
+    inviteMocked(
+        email: string,
+        organization: IOrganization,
+        role: OrganizationRole,
+        sender: string
+    ): ISuccessResponse {
         this.invitationCounter++;
 
         const organizationInvitation: IOrganizationInvitation = {
             id: this.invitationCounter,
             email,
             role,
-            organization: { id: organizationId } as IOrganization,
-            status: OrganizationInvitationStatus.Pending
+            organization,
+            status: OrganizationInvitationStatus.Pending,
+            sender
         };
 
         this.invitationStorage.set(organizationInvitation.id, organizationInvitation);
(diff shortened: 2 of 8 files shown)


packages/origin-backend-core/src/OrganizationInvitation.ts
```diff
@@ -20,11 +20,9 @@ export interface IOrganizationInvitationProperties {
 }
 
 export interface IOrganizationInvitation extends IOrganizationInvitationProperties {
-    organization: IOrganization | IOrganization['id'];
-}
-
-export interface IOrganizationInvitationWithRelations extends IOrganizationInvitation {
     organization: IOrganization;
+    sender: string;
+    processed: boolean;
```
kosecki123: Can we use the status of the invitation to represent if invitation was received or not?

So IMHO adding extra statuses to already existing enum should solve the problem:

OrganizationInvitationStatus.Accepted
OrganizationInvitationStatus.Rejected
OrganizationInvitationStatus.Sent
OrganizationInvitationStatus.Delivered

Is there a use case which requires us to have more than 1 status at time?

packages/origin-backend/migrations/1596616659110-OrganizationInvitationSenderEmail.ts
```diff
@@ -0,0 +1,27 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class OrganizationInvitationSenderAndProcessed1596616659110 implements MigrationInterface {
+    name = 'OrganizationInvitationSenderAndProcessed1596616659110';
+
+    public async up(queryRunner: QueryRunner): Promise<any> {
+        await queryRunner.query(
+            `ALTER TABLE "organization_invitation" ADD "sender" character varying NOT NULL`,
```
kosecki123: Adding NOT NULL columns require a default value or will not migrate already existing DB records and fail the process.

kosecki123 (changes requested):
Please check my comments.

packages/origin-backend-core/src/OrganizationInvitation.ts
```diff
@@ -2,9 +2,10 @@ import { IOrganization } from './Organization';
 import { Role } from './User';
 
 export enum OrganizationInvitationStatus {
-    Pending,
     Rejected,
-    Accepted
+    Accepted,
+    Sent,
+    Viewed
```
kosecki123, line 8: Do not reorder those enums, as existing records in DB will be incorrectly mapped:

Previous Pending = Rejected
Rejected = Accepted ....

So either append only or assign int numbers explicitly so they are not messed up.

kosecki123 (changes requested):
Enum reorder.

packages/origin-backend-core/src/OrganizationInvitation.ts
```diff
@@ -2,9 +2,10 @@ import { IOrganization } from './Organization';
 import { Role } from './User';
 
 export enum OrganizationInvitationStatus {
-    Pending,
-    Rejected,
-    Accepted
+    Rejected = 1,
+    Accepted = 2,
+    Sent = 3,
```
kosecki123: What's the different between Sent and Pending? What happen to existing Pending invitations?

packages/exchange/src/pods/order/order.controller.ts
```diff
@@ -84,7 +84,7 @@ export class OrderController {
     }
 
     @Get()
-    @UseGuards(AuthGuard(), ActiveUserGuard)
+    @UseGuards(AuthGuard())
```
kosecki123: Why?

JGiter: According to https://energyweb.atlassian.net/wiki/spaces/OR/pages/1012989957/Organizations+and+users all users have access to Exchange where orders are listed

kosecki123: /orders endpoint is for managing organization orders. Non active user should not be able to use it.

kosecki123 (approved):

PR #1370, opened 2020-08-07
bug/unable register device

JGiter:
Fixed issue: User belonging to an approved organization cannot register a device with reason that he do not belong to an approved organization


kosecki123:
This good change, would it be possible to remove `IUserWithRealtionsId` ? Also rename IUserWithRelations` to `IUser`? I don't see any purpose of having 2 different interfaces based on that PR, maybe they used somewhere else though.

kosecki123 (approved):

PR #1407, opened 2020-08-18
feature/register platfrom organization form

JGiter:
- Since organization registration data significally changed OrganizationForm component is not suitable now for both registration and viewing and therefore OrganizationForm component replaced by OrganizationView and PlatformOrganizationRegistrationForm components
- In order to provide consistent design and to reuse components for uploading registration documents Upload component was reused which doesn't match upload components on wireframe


packages/origin-ui-core/src/components/Organization/OrganizationView.tsx
```diff
@@ -1,11 +1,228 @@
...
+                        fullWidth
+                    />
+
+                    <TextField
+                        label="CEO passport number"
+                        value={formValues.ceoPassportNumber}
+                        disabled
+                        className="mt-3"
+                        fullWidth
+                    />
```
kosecki123, line 149: Why do we keep those fields? They are not part of the new designs.

JGiter: Its just renamed OrganizationForm which will only be used for viewing organization. I didn't changed it in this PR because PR is only about new organization registration form

kosecki123 (approved):

PR #1424, opened 2020-08-22
feature/IREC account message window

JGiter:
Implemented ConnectOrRegisterIREC registration step.
Added stubs for ConnectIREC, RegisterIREC and RegistrationCompleted registration steps

### DIFF
--- a/packages/localization/src/languages/en.json
+++ b/packages/localization/src/languages/en.json
@@ -492,7 +492,14 @@
                 "signatoryAddress": "Signatory Address",
                 "signatoryEmail": "Singatory Email",
                 "signatoryTelephone": "Signatory Telephone",
-                "uploadSignatoryId": "Upload Singatory ID"
+                "uploadSignatoryId": "Upload Singatory ID",
+                "titleConnectOrRegisterIREC": "Thank you for registering organization on TK marketplace!",
+                "contentConnectOrRegisterIREC": "We are checking your information as soon as possible and will contact you once everything is approved and you can start trading.<1/> <2/> In order to register devices and request I-RECs, users also need to connect to existing I-REC account.",
+                "actions":{
+                    "notNow": "Not now",
+                    "connectIREC": "Connect existing I-REC account",
+                    "registerIREC": "Register new I-REC account"
+                }
             }
         },
         "bundle": {
--- a/packages/origin-ui-core/src/components/Organization/IRECConnectOrRegisterStep.tsx
+++ b/packages/origin-ui-core/src/components/Organization/IRECConnectOrRegisterStep.tsx
@@ -1,0 +1,78 @@
+import React from 'react';
+import { Paper, Grid, Box, Button, Theme, useTheme } from '@material-ui/core';
+import { STEP_NAMES } from './OrganizationRegistrationStepper';
+import OrgAddedIcon from '../../../assets/icon-org-added.svg';
+import { useTranslation } from '../..';
+import { Trans } from 'react-i18next';
+import { useLinks } from '../../utils';
+import { useSelector } from 'react-redux';
+import { getUserOffchain } from '../../features/users/selectors';
+import { useHistory } from 'react-router-dom';
+
+interface IProps {
+    nextStep: (step: STEP_NAMES) => void;
+}
+
+export const IRECConnectOrRegisterStep = (props: IProps) => {
+    const { nextStep } = props;
+    const { t } = useTranslation();
+    const { spacing }: Theme = useTheme();
+    const { getOrganizationViewLink } = useLinks();
+    const orgId = useSelector(getUserOffchain)?.organization?.id;
+    const history = useHistory();
+
+    return (
+        <Paper elevation={1} style={{ padding: spacing(2) }}>
+            <Grid container direction="column">
+                <Grid item container>
+                    <Grid item xs={3}>
+                        <img src={OrgAddedIcon} style={{ maxWidth: '100%', maxHeight: '100%' }} />
+                    </Grid>
+                    <Grid item xs={9}>
+                        <Box fontWeight="fontWeightBold">
+                            {t('organization.registration.titleConnectOrRegisterIREC')}
+                        </Box>
+                        <br />
+                        <Box style={{ fontWeight: 'lighter' }}>
+                            <Trans i18nKey="organization.registration.contentConnectOrRegisterIREC">
+                                We are checking your information as soon as possible and will
+                                contact you once everything is approved and you can start trading.
+                                <br /> <br /> In order to register devices and request I-RECs, also
+                                need to connect to existing I-REC account
+                            </Trans>
+                        </Box>
+                    </Grid>
+                </Grid>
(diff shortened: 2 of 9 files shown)


packages/localization/src/languages/en.json
```diff
@@ -492,7 +492,14 @@
                 "signatoryAddress": "Signatory Address",
                 "signatoryEmail": "Singatory Email",
                 "signatoryTelephone": "Signatory Telephone",
-                "uploadSignatoryId": "Upload Singatory ID"
+                "uploadSignatoryId": "Upload Singatory ID",
+                "titleConnectOrRegisterIREC": "Thanks for registering organization on TK marketplace!",
+                "contentConnectOrRegisterIREC": "We are checking your information as soon as possible and will contact you once everything is approved and you can start trading.<1/> <2/> In order to register devcices and request I-RECs, users also need to connect to existing I-REC account.",
+                "actions":{
+                    "notNow": "Not now",
+                    "connectIREC": "Connect existting I-REC account",
```
josipbagaric: Typo `existing`.

packages/localization/src/languages/en.json
```diff
@@ -492,7 +492,14 @@
                 "signatoryAddress": "Signatory Address",
                 "signatoryEmail": "Singatory Email",
                 "signatoryTelephone": "Signatory Telephone",
-                "uploadSignatoryId": "Upload Singatory ID"
+                "uploadSignatoryId": "Upload Singatory ID",
+                "titleConnectOrRegisterIREC": "Thanks for registering organization on TK marketplace!",
```
josipbagaric: Formal `Thank you` instead of informal `Thanks`.

packages/origin-ui-core/src/components/Organization/IRECConnectOrRegisterStep.tsx
```diff
@@ -0,0 +1,71 @@
...
+                        <Box fontWeight="fontWeightBold">
+                            {t('organization.registration.titleConnectOrRegisterIREC')}
+                        </Box>
+                        <br />
+                        <Box style={{ fontWeight: 'lighter' }}>
+                            <Trans i18nKey="organization.registration.contentConnectOrRegisterIREC">
+                                We are checking your information as soon as possible and will
+                                contact you once everything is approved and you can start trading.
+                                <br /> <br /> In order to register devcices and request I-RECs,
+                                users also need to connect to existing I-REC account
```
josipbagaric: Can't we default to English instead of defining this string twice?

packages/origin-ui-core/src/components/Organization/PlatformOrganizationRegistrationForm.tsx
```diff
@@ -84,7 +84,13 @@ const VALIDATION_SCHEMA = Yup.object({
...
 });
 
-export function PlatformOrganizationRegistrationForm() {
+interface IProps {
+    nextStep: (step: STEP_NAMES) => void;
+}
+
+export function PlatformOrganizationRegistrationForm(props: IProps) {
+    const { nextStep } = props;
+    /* eslint-disable @typescript-eslint/no-unused-vars */
```
josipbagaric, line 93: Trying to understand why this is needed? Don't see any new unused variables.

If it's needed for the whole file, should be put on the top of the file.
If it's only for one line, should be used as `// eslint-disable-next-line @typescript-eslint/no-unused-vars`

packages/localization/src/languages/en.json
josipbagaric: ~Also, what's a `TK marketplace`?~
Nevermind, I see it's a placeholder.

packages/origin-ui-core/src/components/Organization/IRECConnectOrRegisterStep.tsx
JGiter: Not sure how to avoid this duplication. In the documentation there are two ways of providing default values - by using children or default property. Could you point out how this can be done?

packages/origin-ui-core/src/components/Organization/PlatformOrganizationRegistrationForm.tsx
JGiter: Its temporarily until new organization registration endpoint is ready

packages/origin-ui-core/src/components/Organization/IRECConnectOrRegisterStep.tsx
josipbagaric: I thought it already automatically defaults to English if you use `t('organization.registration.contentConnectOrRegisterIREC')`. Could you check if it does or doesn't?

packages/localization/src/languages/en.json
```diff
@@ -492,7 +492,14 @@
...
+                "contentConnectOrRegisterIREC": "We are checking your information as soon as possible and will contact you once everything is approved and you can start trading.<1/> <2/> In order to register devcices and request I-RECs, users also need to connect to existing I-REC account.",
```
josipbagaric: Typo `devices` instead of `devcices`.

packages/origin-ui-core/src/components/Organization/IRECConnectOrRegisterStep.tsx
JGiter: The problem is that I also use this child element to interpolate the html nodes. The documentation mentions that it can also be done through `components` property, but when I try to list nodes like `components = {[', <br />, <br />]}` eslint complains that the `Missing" key "prop for element in array`. May be `components` property should be used in some other way, but the documentation on this topic is very sparse.

josipbagaric: If I understand it correctly, the problem is that we're using HTML elements in the translation e.g. `<br />`?

HTML shouldn't be a part of the translation JSON files.

Why not split the translation into 2 different translation strings, and then do `<br />` only in React? (or better yet, use margins and paddings instead of `<br />`).

JGiter: The initial source of confusion is the need to create multiple keys in the JSON file if the text contains HTML nodes. react i18next solves this problem using the Trans component and for this it requires to list the node indices in the JSON file

JGiter: Do you think that it is better to have multiple keys for logically one text than to duplicate translation?

josipbagaric: Not sure. For e.g. now we have a typo that needs to be fixed in 2 different places because the strings are duplicated.
We can leave it as-is, just make sure to fix the typo in both places.

If there's no good solution here, let's then leave this be. 👍

JGiter: Yes, the optimal solution would be to have translation string only in JSON and in Trans only list nodes but I can't find way to do this

josipbagaric (approved):

PR #1436, opened 2020-08-26
fix(origin-ui-core)organization details shows the same

alexworker23:
Fixed issue when opening organization details always showed the same organization in the following scenario:

1. Log in as an admin
2. Navigate to all organizations
3. Whichever one you click it will always show "Trader Organization"

### DIFF
--- a/packages/origin-ui-core/src/components/Organization/OrganizationView.tsx
+++ b/packages/origin-ui-core/src/components/Organization/OrganizationView.tsx
@@ -1,9 +1,11 @@
 import React, { useState, useEffect } from 'react';
 import { useSelector } from 'react-redux';
+import { useParams } from 'react-router-dom';
 import { makeStyles, createStyles, useTheme, Paper, Grid, TextField } from '@material-ui/core';
 import { Skeleton } from '@material-ui/lab';
 import { Countries } from '@energyweb/utils-general';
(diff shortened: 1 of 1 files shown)


packages/origin-ui-core/src/components/Organization/OrganizationView.tsx
```diff
@@ -43,22 +46,38 @@ export function OrganizationView() {
...
+            const getOrganization = async () => {
+                const organization = await organizationClient.getById(parseInt(params.id));
+                if (organization) {
+                    const activeCountriesParsed: number[] = JSON.parse(
+                        organization.activeCountries
+                    );
+                    setValues(organization, activeCountriesParsed);
+                }
+            };
+            getOrganization();
```
josipbagaric: Looks good from the logic perspective, but isn't this a one-line fix?

Changing 
```
const organization = userOffchain?.organization;
```
to:
```
const organization = params.id ? await organizationClient.getById(parseInt(params.id)) : userOffchain?.organization;
```
would avoid code duplication in `useEffect()`.

josipbagaric (approved):

josipbagaric:
@alexworker23 there are some linting errors. Try running `yarn lint-fix` from the repo root.

PR #1439, opened 2020-08-26
bug: migrate to ts 4

JGiter:
Fixed issue: origin-ui-core stopped compiling because the new version of typescript requires typing of custom theme variables


kosecki123:
It's good but maybe update the typescript libraries so we are 100% sure it passes. Or even better push this commit to already opened typescript upgrade PR.

kosecki123 (changes requested):
As commented above

PR #1440, opened 2020-08-27
Issuer API

josipbagaric:
Moves the **interaction with the smart contracts and certificates** from the frontend (client library) to the **backend (through an API)**.

- [x] Create a NestJS module for interacting with certificates
    - [x] Issuing certificates
    - [x] Transferring certificates
    - [x] Claiming certificates
    - [x] Requesting certificate issuance
    - [x] Approving certificate issuance
    - [x] Revoking certificate issuance
    - [x] Implement private certificates directly in the NestJS module
    - [x] Implement private certification requests directly in the NestJS module
    - [x] Move `validateGenerationPeriod` to the backend instead of facade
- [x] Move `contractsLookup` away from `@energyweb/origin-backend` and into `@energyweb/issuer-api`
- [x] Replace Configuration.Entity with only BlockchainProperties in `@energyweb/issuer`
- [x] Leave `@energyweb/issuer` as a facade for interacting with the blockchain directly (for public certificates only)
- [x] Remove using the `@energyweb/issuer` package from the UI, only interact with certificates through `@energweb/issuer-api`
- [x] Fix showing certificate events

Before merging:
- [x] Squash all migrations in `@energyweb/issuer-api`
- [x] Rebase commits


alexworker23 (approved):

packages/issuer-api-client/src/api.ts
```diff
@@ -0,0 +1,1463 @@
+// tslint:disable
```
kosecki123: Does the openapi generator tool provides any guidelines for storing those generated files? Typical approach is to keep autogenerated sources outside of git repository.

packages/issuer-api/package.json
```diff
@@ -0,0 +1,101 @@
...
+        "typeorm": "ts-node -r tsconfig-paths/register ./node_modules/.bin/typeorm --config ormconfig-dev.ts",
+        "typeorm:migrate": "yarn typeorm migration:generate -- -n",
+        "typeorm:run": "yarn typeorm migration:run",
+        "typeorm:drop": "yarn typeorm schema:drop",
+        "typeorm:dropAndMigrate": "yarn typeorm:drop && yarn typeorm:run",
+        "precommit": "lint-staged"
+    },
+    "dependencies": {
+        "@energyweb/issuer": "2.4.6",
+        "@energyweb/origin-backend": "8.0.0",
```
kosecki123: We should make sure that Issuer package runs without `origin-backend` dependency.

packages/issuer-api/src/pods/certification-request/commands/create-certification-request.dto.ts
```diff
@@ -0,0 +1,24 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class ICreateCertificationRequestDTO {
```
kosecki123: Looks weird, typically we use `I` prefix with interfaces not classes.

packages/issuer-api/src/pods/certification-request/handlers/approve-certification-request.handler.ts
```diff
@@ -0,0 +1,71 @@
...
+    async execute(command: ApproveCertificationRequestCommand): Promise<ISuccessResponse> {
+        const { id } = command;
+
+        const { requestId, isPrivate, energy, approved, owner } = await this.repository.findOne(id);
+
+        if (approved) {
+            throw new BadRequestException({
+                success: false,
+                message: `Certificate #${id} has already been approved`
+            });
```
kosecki123: We should avoid using Http specific exceptions from handlers and services, this make re-use with other tech like messaging hard.

packages/issuer-api/src/pods/certification-request/handlers/create-certification-request.handler.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        const blockchainProperties = await this.blockchainPropertiesService.get();
+
+        const certReq = await CertificationRequestFacade.create(
+            fromTime,
+            toTime,
+            deviceId,
+            blockchainProperties.wrap(),
+            to
+        );
```
kosecki123, line 36: Where the validation for all those inputs is performed?

packages/issuer-api/src/pods/certificate/listeners/on-chain-certificates.listener.ts
```diff
@@ -0,0 +1,102 @@
...
+
+        this.provider.on(
+            this.registry.filters.TransferSingle(null, null, null, null, null),
+            (event: providers.Log) => this.processEvent(EventType.TransferSingle, event)
+        );
+
+        this.provider.on(
+            this.registry.filters.ClaimSingle(null, null, null, null, null, null),
+            (event: providers.Log) => this.processEvent(EventType.ClaimSingle, event)
+        );
```
kosecki123, line 54: Bulk transfers are missing.

packages/issuer-api/test/database.service.ts
```diff
@@ -0,0 +1,24 @@
...
+                .executeForPromise(() => this.connection.query(`TRUNCATE ${tables} CASCADE;`));
+        } catch (error) {
+            throw new Error(`ERROR: Cleaning test db: ${error}`);
+        }
+    }
+
+    public async truncate(table: string) {
+        return this.connection.query(`TRUNCATE "${table}" CASCADE;`);
+    }
+}
```
kosecki123: This file is part of `origin-backend-utils` package.

packages/issuer-api/tslint.json
```diff
@@ -0,0 +1,18 @@
...
+    "member-access": [false],
+    "ordered-imports": [false],
+    "max-line-length": [true, 150],
+    "member-ordering": [false],
+    "interface-name": [false],
+    "arrow-parens": false,
+    "object-literal-sort-keys": false
+  },
+  "rulesDirectory": []
+}
```
kosecki123: Should be deleted,  I assume this comes from nest.js default scaffolding.

kosecki123:
`migrations` project Dockerfiles are missing `issuer-api` DB migration:

https://github.com/energywebfoundation/origin/blob/9bcd9e132cd0979c59bf66d0b5d41ff8a167199a/packages/migrations/deployment/migrations/canary/Dockerfile
https://github.com/energywebfoundation/origin/blob/9bcd9e132cd0979c59bf66d0b5d41ff8a167199a/packages/migrations/deployment/migrations/stable/Dockerfile

packages/issuer-api/package.json
josipbagaric: You're right, didn't update the approach once you did the top-level module stuff.

packages/issuer-api-client/src/api.ts
josipbagaric: I'll look into better ways, but I reckon we would then probably need to keep `@energyweb/issuer-api` as the dependency to `@energyweb/origin-ui-core`. Maybe I can keep the `src/` folder empty until the package builds on npm, but I'll see.

packages/issuer-api/src/pods/certification-request/commands/create-certification-request.dto.ts
josipbagaric: Good catch. Used to be an interface, but had to turn it into a class to show up in the autogenerated client.

packages/issuer-api/tslint.json
josipbagaric: Probably, @energyweb/exchange and irec api also has it.

packages/issuer-api-client/src/api.ts
kosecki123: Having the dependency of `@energyweb/issuer-api` to `@energyweb/origin-ui-core` would be bad, I think the main reason of having the issuer api client is the decouple these two.

josipbagaric: Yes, we definitely need to avoid that.

packages/issuer-api-client/package.json
```diff
@@ -0,0 +1,49 @@
...
+        "client:clean": "find src/ -type f  ! -name \"*.ts\"  -delete"
+    },
+    "dependencies": {
+        "@energyweb/issuer-api": "0.0.1",
+        "@nestjs/testing": "7.4.4",
+        "@nestjs/swagger": "4.6.1",
+        "@nestjs/typeorm": "7.1.4",
+        "@openapitools/openapi-generator-cli": "1.0.18-4.3.1",
+        "axios": "0.20.0",
+        "json-to-pretty-yaml": "1.2.2"
```
kosecki123: Since the generation is now happening on the build stage, should we move those dependencies to the `devDependencies` (at least most of them I guess) ?

kosecki123 (approved):

PR #1443, opened 2020-08-27
feature/login when no invitations

JGiter:
Цhen a user who is not invited to any of the organizations login in for the first time a modal dialog window appears where user proposed to register organization

image

### DIFF
--- a/packages/localization/src/languages/en.json
+++ b/packages/localization/src/languages/en.json
@@ -473,6 +473,8 @@
                 }
             },
             "registration": {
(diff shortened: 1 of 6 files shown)


packages/localization/src/languages/en.json
```diff
@@ -473,6 +473,8 @@
                 }
             },
             "registration": {
+                "loginNoInvitationsModalTitle": "Thanks for registering as a user on the [platform name] marketplace",
```
josipbagaric: Formal `Thank you` instead of informal `Thanks`.

josipbagaric (approved):

PR #1457, opened 2020-08-29
fix/remove inappropriate actions of processed invitation ov 258

alexworker23:
Remove inappropriate actions of processed invitation.
After invitation is accepted or rejected Accept or Reject actions are still appears in organization/invitations tab: 
image-20200821-103259

Done:
1.  
1 BothPending

2. 
2  OneWouldAccept

3.
3  AfterAccepted_OnePending 
(This was fixed in the second commit)

4.
4  OneAcceptedOneRejected

### DIFF
--- a/packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
+++ b/packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
@@ -146,22 +146,6 @@ export function OrganizationInvitationTable(props: IProps) {
         dispatch(setLoading(false));
     }
 
-    const actions =
-        typeof props.organizationId === 'undefined'
-            ? [
-                  {
-                      icon: <Check />,
-                      name: 'Accept',
-                      onClick: (row: string) => accept(parseInt(row, 10))
-                  },
-                  {
-                      icon: <Clear />,
-                      name: 'Reject',
-                      onClick: (row: string) => reject(parseInt(row, 10))
-                  }
-              ]
-            : [];
-
     const columns = [
         { id: 'organization', label: 'Organization' },
         { id: 'email', label: 'Email' },
@@ -175,6 +159,27 @@ export function OrganizationInvitationTable(props: IProps) {
             email: invitation.email
         };
     });
+
+    const actions =
+        typeof props.organizationId === 'undefined'
+            ? rows.map((currentRow) =>
+                  OrganizationInvitationStatus[currentRow.status] ===
+                  OrganizationInvitationStatus.Pending
+                      ? [
+                            {
+                                icon: <Check />,
+                                name: 'Accept',
+                                onClick: (row: string) => accept(parseInt(row, 10))
+                            },
+                            {
+                                icon: <Clear />,
+                                name: 'Reject',
+                                onClick: (row: string) => reject(parseInt(row, 10))
+                            }
+                        ]
+                      : []
+              )
+            : [];
 
     return (
         <TableMaterial
--- a/packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
+++ b/packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
@@ -16,6 +16,7 @@ import {
 } from '../Table/PaginatedLoaderHooks';
 import { getOffChainDataSource } from '../../features/general/selectors';
 import { refreshUserOffchain } from '../../features/users/actions';
+import { useTranslation } from '../..';
 
 interface IRecord {
     organization: IOrganization;
@@ -43,6 +44,7 @@ export function OrganizationInvitationTable(props: IProps) {
     const organizationClient = useSelector(getOffChainDataSource)?.organizationClient;
 
     const dispatch = useDispatch();
+    const { t } = useTranslation();
 
     async function getPaginatedData({
         requestedPageSize,
@@ -172,12 +174,12 @@ export function OrganizationInvitationTable(props: IProps) {
                       ? [
                             {
                                 icon: <Check />,
-                                name: 'Accept',
+                                name: t('organization.invitations.actions.accept'),
                                 onClick: (row: string) => accept(parseInt(row, 10))
                             },
                             {
                                 icon: <Clear />,
-                                name: 'Reject',
+                                name: t('organization.invitations.actions.decline'),
                                 onClick: (row: string) => reject(parseInt(row, 10))
                             }
                         ]
--- a/packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
+++ b/packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
@@ -105,6 +105,10 @@ export function OrganizationInvitationTable(props: IProps) {
 
         try {
             await organizationClient.acceptInvitation(invitation.id);
+            const invitations = await organizationClient.getInvitationsForEmail(props.email);
+            invitations
+                .filter((inv) => inv.id !== invitation.id)
+                .forEach((inv) => organizationClient.rejectInvitation(inv.id));
 
             showNotification(`Invitation accepted.`, NotificationType.Success);
             dispatch(refreshUserOffchain());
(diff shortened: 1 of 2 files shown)


packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
```diff
@@ -90,6 +90,10 @@ export function OrganizationInvitationTable(props: IProps) {
 
     async function accept(rowIndex: number) {
         const invitation = paginatedData[rowIndex]?.invitation;
+        const invitations = await organizationClient.getInvitationsForEmail(props.email);
+        invitations
+            .filter((inv) => inv.id !== invitation.id)
+            .forEach((inv) => organizationClient.rejectInvitation(inv.id));
 
         if (
```
JGiter: Invitation status must be checked before other invitations are rejected

packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
```diff
@@ -90,6 +90,10 @@ export function OrganizationInvitationTable(props: IProps) {
 
     async function accept(rowIndex: number) {
         const invitation = paginatedData[rowIndex]?.invitation;
+        const invitations = await organizationClient.getInvitationsForEmail(props.email);
+        invitations
+            .filter((inv) => inv.id !== invitation.id)
+            .forEach((inv) => organizationClient.rejectInvitation(inv.id));
```
josipbagaric: Did I understand this correctly:
When you accept an organization invite, all other organization invitations are rejected?

packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
```diff
@@ -176,6 +164,26 @@ export function OrganizationInvitationTable(props: IProps) {
         };
     });
 
+    const actions =
+        typeof props.organizationId === 'undefined'
+            ? rows.map((currentRow) =>
+                  currentRow.status === 'Pending'
```
josipbagaric: Please compare with enumeration `OrganizationInvitationStatus` instead of a string `'Pending'`.

e.g. `OrganizationInvitationStatus[currentRow.status] === OrganizationInvitationStatus.Pending`.

packages/origin-ui-core/src/components/Table/TableMaterial.tsx
```diff
@@ -64,7 +64,7 @@ interface IProps<T extends readonly ITableColumn[]> {
     loadPage?: (page: number, filters?: ICustomFilter[]) => void | Promise<any>;
     pageSize?: number;
     total?: number;
-    actions?: ITableAction[];
+    actions?: ITableAction[] | ITableAction[][];
```
JosephBagaric, line 67: I'm trying to understand why we have a 2D array of `ITableAction`. What's the reason for it?

packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
alexworker23: Yes

packages/origin-ui-core/src/components/Table/TableMaterial.tsx
josipbagaric, line 67: ~I'm trying to understand why we need a 2D array of ITableAction. What's the reason for it?~

Nevermind I got it, it's because we now separate actions by row?

alexworker23: Yes, that is the reason

packages/origin-ui-core/src/components/Table/TableMaterial.tsx
```diff
@@ -121,6 +121,57 @@ export function TableMaterial<T extends readonly ITableColumn[]>(props: IProps<T
...
+                            id={id}
+                        />
+                    </TableCell>
+                );
+            }
+            if (arrayCheck && actionsArr[rowId].length === 0) {
+                return <TableCell key={id}></TableCell>;
+            }
+        }
+    }
```
josipbagaric, line 162: This method is way too complex and ambiguous, took me like a minute to understand what it does.

Consider simplifying the implementation and making it more readable.
e.g. Name the variable `is2DArray` instead of `arrayCheck`.

Quick sketch how I would simplify it while maintaining readability. If you can simplify it even further, the better.

```
if (!actionsArr?.length) {
    return <TableCell key={id}></TableCell>;
}

const is2DArray = Array.isArray(actionsArr[0]);

let finalActionsList;

if (!is2DArray) {
    finalActionsList = actionsArr;
} else if (actionsArr[rowId].length > 0) {
    finalActionsList = actionsArr[rowId];
} else {
    return <TableCell key={id}></TableCell>;
}

return (
    <TableCell key={id} className={classes.tableCellWrappingActions}>
        <Actions
            actions={
                allowed
                    ? finalActionsList.filter((action) =>
                            allowed[(row as any).source]?.includes(action.id)
                        )
                    : finalActionsList
            }
            id={id}
        />
    </TableCell>
);
```

packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
alexworker23: Ok

packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
josipbagaric: Looks weird to me. `organizationClient.rejectInvitation()` is an async method, but I see no Promises being resolved here.
Does it work properly?

josipbagaric: Now when I think about it, this logic should be moved to the backend.

josipbagaric: It should also happen after an invitation has been accepted, not before like in this code change.

alexworker23: So should I remove this functionality from here?
The same behaviour is also applied in the PendingInvitationsModal window in accept and reject buttons.

kosecki123: Agree, this kind of decisions should not be handled by the UIs but I see this more or less as an workaround on the fact that user cannot participate multiple organizations.

josipbagaric: @alexworker23 Yes please remove it, here and in other places. I can update the backend part to this branch.

josipbagaric: @kosecki123 should we even bother with this on the backend, or will it be refactored?

kosecki123: No, leave it for now. This will be handled be organizations rewrite.

josipbagaric: @kosecki123 OK.
@alexworker23 Leave it in for now

packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
```diff
@@ -176,6 +164,26 @@ export function OrganizationInvitationTable(props: IProps) {
...
+                      ? [
+                            {
+                                icon: <Check />,
+                                name: 'Accept',
```
josipbagaric: Could you also please localize this? Thanks.

packages/origin-ui-core/src/components/Organization/OrganizationInvitationTable.tsx
```diff
@@ -176,6 +164,26 @@ export function OrganizationInvitationTable(props: IProps) {
...
+                                onClick: (row: string) => accept(parseInt(row, 10))
+                            },
+                            {
+                                icon: <Clear />,
+                                name: 'Reject',
```
josipbagaric: Could you also please localize this? Thanks.

JGiter (changes requested):

josipbagaric (approved):

JGiter (approved):

PR #1471, opened 2020-09-02
fix/configure access according to permissions

alexworker23:
Сonfigure access according to permissions

### DIFF
--- a/packages/origin-ui-core/src/components/Organization/Organization.tsx
+++ b/packages/origin-ui-core/src/components/Organization/Organization.tsx
@@ -5,7 +5,7 @@ import { Role, isRole, UserStatus } from '@energyweb/origin-backend-core';
 
 import { PageContent } from '../PageContent/PageContent';
 import { useLinks } from '../../utils/routing';
-import { getUserOffchain, getIRECAccount } from '../../features/users/selectors';
+import { getUserOffchain, getIRECAccount, getInvitations } from '../../features/users/selectors';
 import { OrganizationTable } from './OrganizationTable';
 import { OrganizationView } from './OrganizationView';
 import { OrganizationInvite } from './OrganizationInvite';
@@ -22,10 +22,16 @@ export const roleNames = {
 
 export function Organization() {
     const user = useSelector(getUserOffchain);
+    const invitations = useSelector(getInvitations);
+    const showInvitations: boolean =
+        user?.organization?.id && isRole(user, Role.OrganizationAdmin)
+            ? true
+            : invitations.length > 0;
 
     const { getOrganizationLink } = useLinks();
 
     const isLoggedIn = Boolean(user);
+    const userIsActive = user && user.status === UserStatus.Active;
     const organization = useSelector(getUserOffchain)?.organization;
     const irecAccount = useSelector(getIRECAccount);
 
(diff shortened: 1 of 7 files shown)


packages/origin-ui-core/src/components/Organization/Organization.tsx
```diff
@@ -22,10 +22,14 @@ export const roleNames = {
 
 export function Organization() {
     const user = useSelector(getUserOffchain);
+    const invitations = useSelector(getInvitations);
+    const showInvitations =
+        user?.organization?.id && isRole(user, Role.OrganizationAdmin) ? true : invitations.length;
```
JGiter: Make showInvitations boolean
Invitations shouldn't be visible to anyone except Organization Admin

packages/origin-ui-core/src/components/Organization/Organization.tsx
```diff
@@ -34,63 +38,60 @@ export function Organization() {
...
             component: OrganizationView,
-            hide: true
+            show: false
         },
         {
             key: 'register-irec',
             label: 'Register I-REC',
             component: IRECRegisterForm,
-            hide: !organization || irecAccount
+            show: organization || irecAccount
```
JGiter: Should be shown if exists organization and no irec

packages/origin-ui-core/src/components/certificates/Certificates.tsx
```diff
@@ -49,19 +49,24 @@ export function Certificates() {
     const isIssuer = isRole(user, Role.Issuer);
     const userIsActive = user && user.status === UserStatus.Active;
 
+    const roleCheck = () =>
```
JGiter: The function checking role should return false if the user is not in the organization. Therefore if in such case isRole returns true then instead of checking role in each component, it is better to do this check in the isRole function
Also roleCheck is too vague name

packages/origin-ui-core/src/components/certificates/Certificates.tsx
```diff
@@ -49,19 +49,24 @@ export function Certificates() {
...
+        userIsActive &&
+        isRole(user, Role.OrganizationUser, Role.OrganizationDeviceManager, Role.OrganizationAdmin);
+
     const CertificatesMenu = [
         {
             key: 'inbox',
             label: 'navigation.certificates.inbox',
             component: InboxCertificates,
-            show: user && !isIssuer,
+            show: isIssuer || roleCheck(),
```
JGiter: Shouldn't inbox be hidden from issuer?

packages/origin-ui-core/src/components/devices/Device.tsx
```diff
@@ -70,55 +73,62 @@ export function Device() {
...
+            show: roleCheck()
         },
         {
             key: 'pending',
             label: t('navigation.devices.pending'),
             component: ProductionPendingList,
             roles: [Role.Issuer],
-            features: [OriginFeature.Devices]
+            features: [OriginFeature.Devices],
+            show: isRole(userOffchain, Role.Issuer)
```
JGiter, line 98: Isn't this a duplication with roles?

packages/origin-ui-core/src/components/devices/Device.tsx
```diff
@@ -70,55 +73,62 @@ export function Device() {
...
+            show: true
         },
         {
             key: 'owned',
             label: t('navigation.devices.my'),
             component: MyDevices,
             roles: [Role.OrganizationDeviceManager, Role.OrganizationAdmin],
-            features: [OriginFeature.Devices, OriginFeature.Seller]
+            features: [OriginFeature.Devices, OriginFeature.Seller],
+            show: roleCheck()
```
JGiter: Isn't this a duplication with roles?

JGiter (changes requested):

packages/origin-ui-core/src/components/Organization/Organization.tsx
alexworker23: Ok

packages/origin-ui-core/src/components/Organization/Organization.tsx
alexworker23: Ok

packages/origin-ui-core/src/components/certificates/Certificates.tsx
alexworker23: Done

JGiter (approved):

josipbagaric (approved):

PR #1475, opened 2020-09-03
fix/configure redirects for logout and accepted invite

alexworker23:
Configure redirects for logout and accepting organization invitation from Invitations table.

1. When logging out, while on some pages that are not allowed for unlogged user - the page can stuck on this page with no content. 

2. When the assigned role for invited user is Member or Device Manager and he accepts the invitation from Invitation table (not from the modal window) he stays in Organizations Tab and can see the pages he is not allowed to see, such as “My organization”, “Invitations”, “Register I-Rec”.


josipbagaric (approved):
LGTM.

PR #1508, opened 2020-09-14
fix/revert role for user removed from organization

alexworker23:
Issue: 
After the user is removed from Organization and his role was Member or Device Manager - he is not able to visit the Organization Tab and register new Organization.


packages/origin-backend/src/pods/organization/organization.service.ts
```diff
@@ -169,6 +169,7 @@ export class OrganizationService {
         }
 
         await this.userService.removeOrganization(memberId);
+        await this.userService.changeRole(memberId, Role.OrganizationAdmin);
```
josipbagaric, line 164: Not sure if this is a proper fix. Users should have no role set if they are not a part of the organization.

Maybe adding a new role `Role.None` would be the way to go, and then adjust the backend endpoints if needed.

kosecki123: This default as OrganizationAdmin is used during user registration as well, it's not the best/cleanest but something we already been relying on.

alexworker23: I agree. This "fix" was done according to the existing roles.

kosecki123 (approved):

josipbagaric (approved):

PR #1512, opened 2020-09-15
feat/ui integration irec register

alexworker23:
This PR provides: 
- IRec Client
- Hide IRec-Connect Features
- Adjusted IRecRegister Form
- FormCountrySelect and MultiCountrySelects now have an optional ISO Format 
- RoleChangedModal moved from AppContainer
- Small fixes to IRec modal windows
- Adjusted localization


packages/origin-ui-core/src/components/Form/FormCountrySelect.tsx
```diff
@@ -9,13 +9,24 @@ interface IFormCountrySelectProps {
...
+    code: country.code
+}));
+
 export function FormCountrySelect(props: IFormCountrySelectProps) {
-    return <FormSelect options={COUNTRY_OPTIONS} {...props} />;
+    if (props.isoFormat) {
+        return <FormSelect options={COUNTRY_OPTIONS_ISO} {...props} />;
+    } else {
+        return <FormSelect options={COUNTRY_OPTIONS} {...props} />;
+    }
```
josipbagaric: Can be a one line: 
`return <FormSelect options={props.isoFormat ? COUNTRY_OPTIONS_ISO : COUNTRY_OPTIONS} {...props} />;`

packages/origin-ui-core/src/components/Organization/Organization.tsx
```diff
@@ -85,10 +82,9 @@ export function Organization() {
             key: 'register-irec',
             label: 'Register I-REC',
             component: IRECRegisterForm,
-            show: enabledFeatures.includes(OriginFeature.IRec) && organization && !irecAccount
+            show: organization && iRecAccount.length === 0
```
josipbagaric: Will this still be hidden if we disable `OriginFeature.IRecConnect`?

I see that you removed `enabledFeatures.includes(OriginFeature.IRec)`

packages/origin-ui-core/src/components/Organization/OrganizationRegistrationStepper.tsx
```diff
@@ -22,8 +22,8 @@ export const OrganizationRegistrationStepper = () => {
         switch (stepName) {
             case STEP_NAMES.REGISTER_ORGANIZATION:
                 return <PlatformOrganizationRegistrationForm nextStep={setActiveStep} />;
-            case STEP_NAMES.CONNECT_OR_CREATE:
-                return <IRECConnectOrRegisterStep nextStep={(step) => setActiveStep(step)} />;
+            // case STEP_NAMES.CONNECT_OR_CREATE:
+            //     return <IRECConnectOrRegisterStep nextStep={(step) => setActiveStep(step)} />;
```
josipbagaric: Remove comments from the code, we have git history for preserving code.

packages/origin-ui-core/src/components/Organization/Organization.tsx
alexworker23: No, this feature now depends on existing organization and existing irecAccount. 
IRecConnect is hidden with enabledFeatures.

packages/origin-ui-core/src/components/Organization/OrganizationRegistrationStepper.tsx
alexworker23: Kept it this way, because want to ask if i should I delete this file? I don't see any usage of this.

kosecki123 (approved):
LGTM 👍

packages/origin-ui-core/src/components/Organization/OrganizationRegistrationStepper.tsx
josipbagaric: Sure, check if it is used. If it's not, feel free to delete it.

packages/origin-ui-core/src/components/Organization/Organization.tsx
josipbagaric: So what you're saying is that putting `enabledFeatures.includes(OriginFeature. IRecConnect)` here would be redundant because of `iRecAccount.length === 0`.

If we are sure the length will always be 0 when we have  `OriginFeature. IRecConnect` disabled, then I guess it's fine.

josipbagaric (approved):

packages/origin-ui-core/src/components/Organization/Organization.tsx
josipbagaric: Wait, this is the I-REC registration form?

We want to have this under a feature toggle, because we have Origin instances where we don't use I-REC.

alexworker23: Ok, will be done this way.

packages/utils-general/src/OriginFeature.ts
```diff
@@ -8,7 +8,8 @@ export enum OriginFeature {
     Devices = 'devices',
     Certificates = 'certificates',
     CertificationRequests = 'certificationRequests',
-    IRec = 'irec'
+    IRec = 'irec',
+    IRecConnect = 'irecConnect'
```
josipbagaric, line 12: Why do we need to have a separate OriginFeature `IRecConnect`? Isn't IRecConnect already a part of the larger IRec feature?

josipbagaric: To better formulate my question, is there some case where we will have `IRec` turned **OFF**, and `IRecConnect` turned **ON**? I assume not.

alexworker23: Now we don't have the backend part for IRec Connect, because it is not clear how exactly the connection with existing I-Rec account will happen.We have only the "Register new IRec Account" part. That's why we should keep I-Rec connect aside for now.

josipbagaric (approved):

PR #1515, opened 2020-09-16
feat/adjust organization and irec view

alexworker23:
This PR provides: 
- IRec Organization view for Admin and Organization Admin
- Adjusted Organization View

myOrg

adminOrgs

### DIFF
--- a/packages/origin-organization-irec-api/src/registration/registration.controller.ts
+++ b/packages/origin-organization-irec-api/src/registration/registration.controller.ts
@@ -7,7 +7,9 @@ import {
     Get,
     Post,
     UseGuards,
-    UseInterceptors
+    UseInterceptors,
+    Param,
+    ParseIntPipe
 } from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
 import { RegistrationDTO } from './registration.dto';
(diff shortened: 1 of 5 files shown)


packages/origin-organization-irec-api/src/registration/registration.controller.ts
```diff
@@ -29,6 +31,17 @@ export class RegistrationController {
...
+    @UseGuards(AuthGuard())
+    public getRegistrationsById(
+        @Param('id', new ParseIntPipe()) orgId: number,
+        @UserDecorator() loggedInUser: LoggedInUser
+    ): Promise<Registration[]> {
+        const isAdmin = loggedInUser.hasRole(Role.Admin, Role.SupportAgent);
+
+        return this.registrationService.find(isAdmin ? orgId.toString() : null);
+    }
+
```
kosecki123: I'm not sure what is the intention of this method, but from what I see it alls admins to get registrations by organization Id and others to get all registrations (which is an obvious security risk).

This is actually the same as `public getRegistrations(@UserDecorator() loggedInUser: LoggedInUser)` above. 
The GET above will return all registrations for all organizations - which is fine - filter by org locally, there isn't many of those registrations expected to be, it's actually 1 by organization.

packages/origin-ui-core/src/utils/irec/IRecClient.ts
```diff
@@ -25,6 +26,13 @@ export class IRecClient implements IIRecClient {
         return response.data;
     }
 
+    public async getRegistrationsById(id: number): Promise<Registration[]> {
+        const response = await this.requestClient.get<unknown, Registration[]>(
+            `${this.registrationEndpoint}/${id}`
+        );
+        return response.data;
+    }
```
kosecki123: To be removed, as commented previously.

kosecki123 (changes requested):
Remove redundant getRegistrationsById method.

kosecki123 (approved):

PR #1544, opened 2020-09-25
feat/document download

alexworker23:
- Send documents with organization data when registering an Organization
- Show attached documents in Organization view for OrganizationAdmin and Platform Admin
- Download documents 
orgViewNoIrec
orgViewWithIRec
severalDocs

### DIFF
--- a/packages/origin-ui-core/src/components/Organization/DownloadDocuments.tsx
+++ b/packages/origin-ui-core/src/components/Organization/DownloadDocuments.tsx
@@ -42,7 +42,6 @@ export const DownloadDocuments = ({ documents, name }) => {
             try {
                 const response = await filesClient.download(id);
                 if (response) {
-                    console.log(response);
                     const imageType = response.headers['content-type'];
                     const blob = new Blob([response.data], { type: imageType });
                     const url = window.URL.createObjectURL(blob);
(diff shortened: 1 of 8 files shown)


kosecki123:
Are these items labeled always as proof and signatory id?

Do we support multiple files for each? And how then this is going to be presented?

alexworker23:
These items always labeled as Company Proof and Signatory ID on the Organization View, and these names are inherited when you download them.

Yes, multiple files are supported for each row. If there are multiple files for one document type they will appear one after another with indexes, just like Company Proofs on the last picture.

packages/origin-ui-core/src/components/Organization/DownloadDocuments.tsx
```diff
@@ -0,0 +1,79 @@
...
+                    const link = document.createElement('a');
+                    link.href = url;
+                    link.setAttribute('download', name);
+                    link.style.display = 'none';
+                    document.body.appendChild(link);
+                    link.click();
+                    document.body.removeChild(link);
+                    setTimeout(() => {
+                        window.URL.revokeObjectURL(url);
+                    }, 100);
```
kosecki123: I think there is no reason to use the download via XHR call. Just use `getLinks` to generate URL and render them on the document, browser will handle the download.

kosecki123 (approved):

packages/origin-ui-core/src/components/Organization/DownloadDocuments.tsx
kosecki123: This flow is currently impossible as <a> element links are not passing the custom headers due to security concerns.

PR #1613, opened 2020-10-13
feat/repeated purchase

alexworker23:
This PR adds new feature of Repeated Purchase (Demand): 
1. You can post a demand on Exchange.
2. You can view your demands on My Orders. 
3. You can edit demand.
4. You can remove demand.

OneTimeForm
repeatedForm
OrdersTable
demandUpdateModal


packages/exchange/src/pods/demand/demand.controller.ts
```diff
@@ -66,6 +68,19 @@ export class DemandController {
         return demand;
     }
 
+    @Put('/:id/update')
+    @UseGuards(AuthGuard(), ActiveUserGuard)
+    public async update(
```
kosecki123: The basic idea on demand/bids/asks updates is that we want to keep them immutable, once created you you should not be able to edit price, volumes etc. This is a basic requirement in order to achieve the transparency and accounting.

The `update` functionality should be seen as combination of:

- pause + archive
- create

This is analogous to how we are currently handling Orders as well.

Additionally when adding new API functionality, the tests to cover this case are must have.

packages/origin-ui-core/src/components/orders/MyOrders.tsx
```diff
@@ -2,20 +2,29 @@ import React from 'react';
...
+    const demands: Demand[] = useSelector(getDemands).filter(
+        (d) => d.status !== DemandStatus.ARCHIVED
+    );
     const orders: Order[] = useSelector(getOrders).filter(
         (o) => o.status === OrderStatus.Active || o.status === OrderStatus.PartiallyFilled
     );
     const asks = orders.filter((o) => o.side === OrderSide.Ask);
-    const bids = orders.filter((o) => o.side === OrderSide.Bid);
+    const bids = orders.filter(
+        (o) => o.side === OrderSide.Bid && !{}.hasOwnProperty.call(o, 'demandId')
```
kosecki123: Possibly better solution would be to:

`o.demandId === undefined`

or

Maybe even better to create custom Orders object like

```
class Orders extends Array<Order> {
  constructor(orders: Order[]) {
    super(...orders);
  }

  public get bids() {
    return this.filter(o=>{logic here});
  }

  ...
}
```

and keep those filtering there, which improves the reusability etc.

packages/origin-ui-core/src/utils/exchange/math.ts
```diff
@@ -8,3 +8,34 @@ export function calculateTotalPrice(priceInDisplayUnit: string, energyInDisplayU
...
+        const divisor = setDivisor(period);
+
+        if (dayDifference > 0 && parsedVolume > 0) {
+            const numberOfOccurrences = Math.round(dayDifference / divisor) + 1;
+            return parsedVolume * numberOfOccurrences;
+        }
+    } else {
+        return '';
+    }
+}
```
kosecki123: This potentially can lead to a situation where UI would present different outcome than it would be then created the API. We should basically use same code/logic for generation on the UI side.

I suggest we add new endpoint such as to `POST /demand/prepare` that would return a populated demand object with bids etc that can be verified by the user before creating a real demand.

kosecki123 (changes requested):

packages/origin-ui-core/src/utils/exchange/math.ts
alexworker23: Ok, will adjust code according to comments

kosecki123 (approved):

PR #1671, opened 2020-11-05
Exchange ui decouple

alexworker23:
In this PR new package is created: **exchange-ui-core**.
Package consists of containers, components, features, styles and utils which are used for creation of Exchange UI.

**origin-ui** package is also changed, and now it contains functionality allowing this package to be the root of the whole UI.

Exchange components are removed from **origin-ui-core** alongside with unused components.


kosecki123:
I think we should merge the commits into `origin-ui`, `exchange-ui-core` and `origin-ui`. These commits are correct from the PR point of view only, but when seen from changelog items for release point of view they not provide any good information.

packages/exchange-ui-core/env.sh
```diff
@@ -0,0 +1,51 @@
...
+  BACKEND_URL
+  BACKEND_PORT
+  BLOCKCHAIN_EXPLORER_URL
+  REGISTRATION_MESSAGE_TO_SIGN
+  ISSUER_ID
+  DEVICE_PROPERTIES_ENABLED
+  DEFAULT_ENERGY_IN_BASE_UNIT
+  EXCHANGE_WALLET_PUB
+  GOOGLE_MAPS_API_KEY
+)
```
kosecki123: Here you probably need only EXCHANGE_WALLET_PUB only

alexworker23:
> I think we should merge the commits into `origin-ui`, `exchange-ui-core` and `origin-ui`. These commits are correct from the PR point of view only, but when seen from changelog items for release point of view they not provide any good information.

You mean to make 3 commits in total,one per package, or to change commits order by package?

kosecki123 (commented):
> > I think we should merge the commits into `origin-ui`, `exchange-ui-core` and `origin-ui`. These commits are correct from the PR point of view only, but when seen from changelog items for release point of view they not provide any good information.
> 
> You mean to make 3 commits in total,one per package, or to change commits order by package?

Make 3 commits.

packages/exchange-ui-core/env.sh
josipbagaric: Maybe DEFAULT_ENERGY_IN_BASE_UNIT also?

josipbagaric (approved):

kosecki123 (approved):

PR #1698, opened 2020-11-11
fix(issuer-api): Return correct HTTP status codes

josipbagaric:
Fixes an issuer where the API would respond with a status 200 even if the action failed.


packages/issuer-api/src/pods/certificate/certificate.controller.ts
```diff
@@ -30,13 +30,16 @@ import { CertificateEvent } from '../../types';
 import { GetAllCertificateEventsQuery } from './queries/get-all-certificate-events.query';
 import { CertificateDTO } from './certificate.dto';
 import { SuccessResponseDTO } from '../../utils/success-response.dto';
+import { ExceptionController } from '../../utils/ExceptionController';
 
 @ApiTags('certificates')
 @Controller('certificate')
-export class CertificateController {
+export class CertificateController extends ExceptionController {
```
kosecki123: I guess ExceptionController  should be just utils function, would be more re-usable etc

kosecki123: Alternatively, you could implement a interceptor to transform return into errors if necessary for e.g like https://github.com/energywebfoundation/origin/blob/d1025d2b034f317520de03866ff4b76bb62c333a/packages/origin-backend-utils/src/interceptor/NullOrUndefinedResultInterceptor.ts

josipbagaric: Good idea, interceptor is a better solution.

ioncreature (approved):

kosecki123 (approved):

PR #1723, opened 2020-11-18
Styles inheritance

alexworker23:
This PR provides: 
- Usage of colors extracted from configuration instead of using scss variables
- Unused scss variables removed
- Login page background is added to Context in order to be configurable


kosecki123:
Would it make sense to scan through commits and try to rebase them? Also, are any of these commits breaking changes?

kosecki123:
Shall we add a wiki page explaining how to use those new variables and what is the outcome?

kosecki123 (approved):

alexworker23:
> Would it make sense to scan through commits and try to rebase them? Also, are any of these commits breaking changes?

1. Yes 
2. No

alexworker23:
> Shall we add a wiki page explaining how to use those new variables and what is the outcome?

Yes, I think I can document it for comfy usage, and also write the rules how to properly use colors in newly created components, so that we can avoid issues with broken color schemes.

kosecki123:
> > Shall we add a wiki page explaining how to use those new variables and what is the outcome?
> 
> Yes, I think I can document it for comfy usage, and also write the rules how to properly use colors in newly created components, so that we can avoid issues with broken color schemes.

Please add documentation here https://github.com/energywebfoundation/origin/wiki/UI-Customizations

kosecki123:
Looks good now 👍

PR #1724, opened 2020-11-18
feat(exchange): separate API for account creation

kosecki123:
This PR makes the functionality of account creation and account balance as two separate API endpoints. 

BREAKING CHANGE: Getting account using GET /account does not longer create a on-chain account automatically. Explicity POST /account has to be used

BREAKING CHANGE: GET /account does not longer returns the balance information, functionality is now availble under GET /account-balance endpoint

UI-PART:
ex-dep-add1
ex-dep-add2
ex-dep-add-3


packages/exchange/src/pods/account/account.controller.ts
```diff
@@ -12,30 +13,58 @@ import {
...
+        const account = await this.accountService.getAccount(ownerId);
+
+        const response = account || { address: '' };
+
+        return response;
+    }
+
+    @Post()
+    @UseGuards(AuthGuard(), ActiveUserGuard, ActiveOrganizationGuard)
+    @ApiResponse({ status: HttpStatus.ACCEPTED, description: 'Created the Exchange Account' })
```
josipbagaric, line 55: This is a bit misleading.

`@Post()` returns 201 or `HttpStatus.CREATED` as the default status.
In the ApiResponse we describe the return status as 202 ACCEPTED

We should either change the description:
`@ApiResponse({ status: HttpStatus.CREATED ...`
or change the actual return code
```
@HttpCode(HttpStatus.ACCEPTED)
@ApiResponse({ status: HttpStatus.ACCEPTED ...
```

packages/exchange/src/pods/account/account.controller.ts
```diff
@@ -12,30 +13,58 @@ import {
...
+    public async create(@UserDecorator() { ownerId }: ILoggedInUser): Promise<boolean> {
+        try {
+            await this.accountService.create(ownerId);
+            return true;
+        } catch (error) {
+            this.logger.error(error.message);
+
+            if (error instanceof AccountAlreadyExistsError) {
+                throw new BadRequestException({ message: error.message });
```
josipbagaric: Would be good to add a Swagger definition for this error e.g.
`@ApiBadRequestResponse({ description: 'Account already exists' })`

Although this looks more like a conflict exception:
```
@ApiConflictResponse({ description: 'Account already exists' })
...
throw new ConflictException({ message: error.message });
```

packages/exchange/src/pods/account/account.controller.ts
```diff
@@ -12,30 +13,58 @@ import {
...
+            }
 
-        return account;
+            throw new InternalServerErrorException({
+                message: `Unable to create an account due an unknown error`
```
josipbagaric, line 69: Maybe add `Unable to create an account due an unknown error ${error.message}` to the message description so that the user can see the actual error message.

kosecki123: This is internal error so nothing usefull for the user.

josipbagaric: You're right. We need to make sure this is logged somewhere on the server then

josipbagaric (approved):

PR #1748, opened 2020-11-26
chore: add missing migrations

josipbagaric:
Adds missing migrations.


kosecki123 (changes requested):
Please use rename

josipbagaric:
> Please use rename

Only found a way to rename the constraints, not change the destination table name of the constraint.

josipbagaric:
But cool thing is this highlighted another bug. The `bundle` table was not renamed to `exchange_bundle`.

kosecki123 (approved):

PR #1761, opened 2020-12-03
feat(origin-backend-utils): Active organization guard

kosecki123:
This PR adds an extra guard that check for organization status to be Active

### DIFF
--- a/packages/origin-backend-utils/src/guard/ActiveOrganizationGuard.ts
+++ b/packages/origin-backend-utils/src/guard/ActiveOrganizationGuard.ts
@@ -1,0 +1,31 @@
+import { IUser, OrganizationStatus } from '@energyweb/origin-backend-core';
+import {
+    CanActivate,
+    ExecutionContext,
+    HttpException,
+    HttpStatus,
+    Injectable
+} from '@nestjs/common';
+
+@Injectable()
+export class ActiveOrganizationGuard implements CanActivate {
+    canActivate(context: ExecutionContext): boolean {
+        const request = context.switchToHttp().getRequest();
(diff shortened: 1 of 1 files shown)


alexworker23 (approved):

packages/origin-backend-utils/src/guard/ActiveOrganizationGuard.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    canActivate(context: ExecutionContext): boolean {
+        const request = context.switchToHttp().getRequest();
+        const user = request.user as IUser;
+        const _user = user as IUser;
+
+        if (_user.organization?.status !== OrganizationStatus.Active) {
+            throw new HttpException(
+                `Only active organizations can perform this action. Your status is ${
+                    OrganizationStatus[_user.status]
```
josipbagaric: Should be `OrganizationStatus[_user.organization.status]` or just `_user.organization.status` since we use string enums.

josipbagaric (changes requested):

packages/origin-backend-utils/src/guard/ActiveOrganizationGuard.ts
```diff
@@ -0,0 +1,25 @@
...
+                }`,
+                412
```
josipbagaric: You can also use `HttpStatus.PRECONDITION_FAILED` instead of `412` to be more explicit.

kosecki123: Changed to Forbidden.

josipbagaric (approved):

PR #1762, opened 2020-12-03
UI fixes and improvements

alexworker23:
In this PR: 
- FIXED:
After selecting a device type during device registration and then removing the selection, the only device type that the user can select is hydro. All other device types cannot be selected anymore. Only after refreshing, can the user choose all device types again. 
- FIXED:
After login User can leave the `PendingInvitationsModal` by clicking on empty space on the page. He should be allowed to leave this modal window only after choosing 1 of the options: Maybe Later, Decline, Accept
- ADDED: 
Warning notification if the User tries to connect the blockchain address which is currently in use by the other User.
- IMPROVED:
Created a separate component for `UpdateSupplyModal`, which previously was a part of `AutoSupplyDeviceTable`.
- FIXED: 
When clicking register on the login page, the link leads to the settings page, not the register user page.
When clicking “connect blockchain address” after the onboarding flow, it leads to the settings page, not the user profile page.
- FIXED:
When selecting the user role in Organization Invitation it would send a string instead of number and it made impossible to create an invite for user.

### DIFF
--- a/packages/origin-ui-core/src/features/users/sagas.ts
+++ b/packages/origin-ui-core/src/features/users/sagas.ts
@@ -229,10 +229,12 @@ function* updateBlockchainAddress(): SagaIterator {
             yield call(callback);
         } catch (error) {
             if (error?.data?.message) {
-                showNotification(error?.data?.message, NotificationType.Error);
+                showNotification(error.data.message, NotificationType.Error);
+            } else if (error?.response) {
+                showNotification(error.response.data.message, NotificationType.Error);
             } else if (error?.message) {
-                console.log(error);
-                showNotification(error?.message, NotificationType.Error);
+                console.log('here it is', error.response.data.message);
+                showNotification(error.message, NotificationType.Error);
             } else {
                 console.warn('Could not log in.', error);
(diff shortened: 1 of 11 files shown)


packages/origin-ui-core/src/features/users/sagas.ts
```diff
@@ -229,10 +229,12 @@ function* updateBlockchainAddress(): SagaIterator {
...
         } catch (error) {
             if (error?.data?.message) {
-                showNotification(error?.data?.message, NotificationType.Error);
+                showNotification(error.data.message, NotificationType.Error);
+            } else if (error?.response) {
+                showNotification(error.response.data.message, NotificationType.Error);
             } else if (error?.message) {
-                console.log(error);
-                showNotification(error?.message, NotificationType.Error);
+                console.log('here it is', error.response.data.message);
```
josipbagaric: Probably not intended?

josipbagaric (approved):

packages/origin-ui-core/src/features/users/sagas.ts
alexworker23: Yes

PR #1766, opened 2020-12-04
chore: store the I-REC connection

josipbagaric:


packages/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
```diff
@@ -0,0 +1,42 @@
...
+        private readonly repository: Repository<Connection>,
+        private readonly registrationService: RegistrationService,
+        private readonly eventBus: EventBus
+    ) {}
+
+    async execute({
+        user: { organizationId },
+        credentials: { userName, password, clientId, clientSecret }
+    }: CreateConnectionCommand): Promise<ConnectionDTO> {
+        const client = new IRECAPIClient(process.env.IREC_API_URL);
```
ioncreature, line 25: @piotr @JosephBagaric Guys, do you have any ideas on how to mock IRECAPIClient for a test?

kosecki123: 1. Make it a dependency instead of creating an instance
2. Use override feature in the test similarly to https://github.com/energywebfoundation/origin/blob/ec5aa08791af0abfe3b1ee059657518dff68c05e/packages/exchange/test/exchange.ts#L146-L147

PR #1779, opened 2020-12-07
Device management overhaul

kosecki123:
This PR adds new API endpoints for handling Origin specific and external issuers device registries.

More details in ADR file doc/architecture/decisions/0012-device-registries.md


packages/origin-device-registry-irec-local-api/src/device/device.module.ts
```diff
@@ -0,0 +1,16 @@
...
+import { DeviceService } from './device.service';
+import { ValidateDeviceOwnershipCommandHandler } from './handlers/validate-device-ownership.handler';
+
+@Module({
+    imports: [TypeOrmModule.forFeature([Device]), CqrsModule],
+    providers: [DeviceService, ValidateDeviceOwnershipCommandHandler],
+    exports: [DeviceService],
+    controllers: [DeviceController]
+})
+export class DeviceModule {}
```
ioncreature, line 16: We already have `DeviceModule` in `origin-backend`. Is it safe to create modules with the same name in Nest?

kosecki123: Yes, since they are located in other packages, you can name them whatever, also the old one will be removed.

packages/origin-device-registry-api/src/device-registry/device.entity.ts
```diff
@@ -0,0 +1,32 @@
...
+    owner: string;
+
+    @Column({ unique: true })
+    externalRegistryId: string;
+
+    @Column({ unique: true })
+    smartMeterId: string;
+
+    @Column('simple-json', { nullable: true })
+    externalDeviceIds?: IExternalDeviceId[];
```
josipbagaric, line 25: Do we still need this when we have externalRegistryId?

packages/origin-device-registry-irec-local-api/src/utils/success-response.dto.ts
```diff
@@ -0,0 +1,14 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { ISuccessResponse } from '@energyweb/origin-backend-core';
+import { IsBoolean, IsString, ValidateIf } from 'class-validator';
+
+export class SuccessResponseDTO implements ISuccessResponse {
```
josipbagaric, line 5: This is used across our system. Maybe we can add this definition to `@energyweb/origin-backend-utils`.

packages/origin-device-registry-irec-local-api/src/device/device.entity.ts
```diff
@@ -0,0 +1,71 @@
...
+
+@Entity({ name: 'irec_device_registry_device' })
+export class Device extends ExtendedBaseEntity {
+    constructor(device: Partial<Device>) {
+        super();
+        Object.assign(this, device);
+    }
+
+    @PrimaryGeneratedColumn('uuid')
+    id: string;
```
josipbagaric, line 13: Can `code` be used as the primary key, instead of generating a new id?
```
@PrimaryColumn('uuid')
code: string;
```

packages/origin-device-registry-api/src/device-registry/device.entity.ts
kosecki123: I'd rather keep it, for some other project we also used real-time data id or something.

packages/origin-device-registry-irec-local-api/src/device/device.entity.ts
kosecki123: Good observation, but in PGSQL the primary key has to be either guid or serial (auto increment), while code is the string.

josipbagaric (approved):

arvindchez:
The PR is very big means the use case can be spilt into smaller jobs? - Just my thoughts

kosecki123:
> The PR is very big means the use case can be split into smaller jobs? - Just my thoughts

Yes, this is another overblown PR, next time will avoid that.

PR #1786, opened 2020-12-09
feat: decouple erc1888 from exchange 

ioncreature:
Decouple Deposit/Withdrawal in the exchange package


packages/exchange-io-erc1888/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "@energyweb/exchange-io-erc1888",
+  "version": "1.0.0",
+  "description": "ERC1888 withdwaral/deposit processing for exchange",
+  "author": "EnergyWeb DevHub GmbH; Aleksandr Marenin, aleksandr.amarenin@energyweb.com",
```
josipbagaric: Is the email correct? :)

packages/exchange-io-erc1888/package.json
```diff
@@ -0,0 +1,72 @@
...
+  "homepage": "https://github.com/energywebfoundation/origin#readme",
+  "license": "ISC",
```
josipbagaric: We release packages under GPLv3. Check other packages for the correct license.

packages/exchange/src/app.module.ts
```diff
@@ -56,8 +54,6 @@ const getEnvFilePath = () => {
         ProductModule,
         AccountDeployerModule,
         AccountBalanceModule,
-        DepositWatcherModule,
-        WithdrawalProcessorModule,
```
josipbagaric, line 60: So Deposits and Withdrawals are disabled in the exchange?
How do we then deposit/withdraw in the future?

ioncreature: We have to import this new module to have erc1888 deposit/withdrawals

packages/exchange-io-erc1888/package.json
ioncreature: fixed ))

packages/exchange-io-erc1888/package.json
ioncreature: done, thanks

packages/exchange-io-erc1888/package.json
josipbagaric: I think it's still incorrect. Should be @enerygweb.org instead of @energyweb.com

ioncreature: 🤦 Thanks, now it is fixed

kosecki123:
This looks good, but commit messages needs rework. Since `exchange` does not have the withdrawals and deposit module anymore, it should be marked as BREAKING CHANGE. 

It's probably 3 commit messages that you need:
- feat(exchange-io-erc1888): withdrawal and deposits for ERC1888 token
- chore(exchange):  withdrawal and deposit removal. BREAKING CHANGE: .....
- chore(origin-backend-app): integration .....

packages/exchange-io-erc1888/test/exchange.ts
```diff
@@ -0,0 +1,198 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
kosecki123, line 1: Seems like this has been copied from exchange module, maybe better import or something rather than duplicating the code?

ioncreature: I copied it and made slight changes, do not see a way to abstract

ioncreature:
> This looks good, but commit messages needs rework. Since `exchange` does not have the withdrawals and deposit module anymore, it should be marked as BREAKING CHANGE.
> 
> It's probably 3 commit messages that you need:
> 
> * feat(exchange-io-erc1888): withdrawal and deposits for ERC1888 token
> * chore(exchange):  withdrawal and deposit removal. BREAKING CHANGE: .....
> * chore(origin-backend-app): integration .....

Done

kosecki123 (approved):

PR #1794, opened 2020-12-18
fix(origin-core-ui): added restriction info for pages requiring excha…

marcin-l-tsh:
added restriction info for pages requiring exchange deposit address


josipbagaric:
@marcin-l-tsh The build seems to be failing. Will review after the build passes.

josipbagaric (approved):

PR #1825, opened 2020-12-21
feat(issuer-api): added validation to "to" address

ioncreature:
Certification request creation "to" address

Certificate request has weak validation and when you send to: "" then it passes but mints to a caller instead of the user account. Since we want to support different send destination types like strings, eth addresses, etc, we should not add @IsEthereumAddress to the validation.

Instead, we should add an @IsNotEmpty() check to the to field in CreateCertificationRequestDTO .

In the CreateCertificationRequestHandler we should then check if this string is an Ethereum address before proceeding with the creation.

### DIFF
--- a/packages/issuer-api/src/pods/certification-request/handlers/create-certification-request.handler.ts
+++ b/packages/issuer-api/src/pods/certification-request/handlers/create-certification-request.handler.ts
@@ -5,18 +5,13 @@ import { InjectRepository } from '@nestjs/typeorm';
 import { Subject } from 'rxjs';
 import { concatMap } from 'rxjs/operators';
 import { Repository } from 'typeorm';
+import { isAddress, getAddress } from 'ethers/lib/utils';
 
 import { BlockchainPropertiesService } from '../../blockchain/blockchain-properties.service';
 import { CertificationRequestStatus } from '../certification-request-status.enum';
 import { CertificationRequestDTO } from '../certification-request.dto';
 import { CertificationRequest } from '../certification-request.entity';
 import { CreateCertificationRequestCommand } from '../commands/create-certification-request.command';
-
-const eth = /^(0x)[0-9a-f]{40}$/i;
-
-function isValidEthereumAddress(to: string) {
-    return typeof to === 'string' && eth.test(to);
-}
 
 @CommandHandler(CreateCertificationRequestCommand)
 export class CreateCertificationRequestHandler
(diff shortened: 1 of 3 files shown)


packages/issuer-api/src/pods/certification-request/handlers/create-certification-request.handler.ts
```diff
@@ -36,6 +42,12 @@ export class CreateCertificationRequestHandler
         files,
         isPrivate
     }: CreateCertificationRequestCommand): Promise<CertificationRequestDTO> {
+        if (!isValidEthereumAddress(to)) {
```
josipbagaric: Please use the latest `ethers.utils.isAddress()` from [ethers](https://docs.ethers.io/v5/single-page/#/v5/api/utils/address/-%23-utils-isAddress) package instead of a custom function, to avoid implementation loopholes.

josipbagaric (changes requested):

packages/issuer-api/src/pods/certification-request/handlers/create-certification-request.handler.ts
```diff
@@ -12,6 +12,12 @@ import { CertificationRequestDTO } from '../certification-request.dto';
 import { CertificationRequest } from '../certification-request.entity';
 import { CreateCertificationRequestCommand } from '../commands/create-certification-request.command';
 
+const eth = /^(0x)[0-9a-f]{40}$/i;
```
JosephBagaric, line 15: This implementation assumes that `a-f` are always lower-case, which is not true for checksummed addresses. Sometimes they are also `A-F`.

josipbagaric:
Please add the following tests:
- Test that passes when a checksummed address is passed (e.g. `0xD173313A51f8fc37BcF67569b463abd89d81844f`)
- Test that passes when a "lowercased" address is passed (e.g. `0xd173313a51f8fc37bcf67569b463abd89d81844f`)

josipbagaric:
IMO we should then cast the address to checksummed, and only store the checksummed address in the backend, and work with checksummed addresses thereafter.

packages/issuer-api/src/pods/certification-request/handlers/create-certification-request.handler.ts
JosephBagaric: Hmm ethers seems to return `false` for a non-checksummed address...

josipbagaric: Actually no need for this, so long as it passes the tests.

ioncreature:
> IMO we should then cast the address to checksummed, and only store the checksummed address in the backend, and work with checksummed addresses thereafter.

Added checksum address saving

ioncreature:
> Please add the following tests:
> 
> * Test that passes when a checksummed address is passed (e.g. `0xD173313A51f8fc37BcF67569b463abd89d81844f`)
> * Test that passes when a "lowercased" address is passed (e.g. `0xd173313a51f8fc37bcf67569b463abd89d81844f`)

Done

josipbagaric (approved):

PR #1829, opened 2020-12-24
chore(exchange-ui-core): show all bids and asks by default

ioncreature:
Now it shows all active bids and asks in ui when no filters specified
image


kosecki123 (changes requested):
As discussed this should be implemented in a way that when selecting a specific Filter like Solar or any other we should also include Bids of type Any in the result.

PR #1831, opened 2020-12-24
fix/increase exchange account creation delay

alexworker23:


josipbagaric:
Is this delay non-blocking? Can the user still interact with the UI during this delay?

alexworker23:
This delay blocks the UI from interaction

josipbagaric:
IMO this delay should be non-blocking. 4 sec is too long for the whole UI to be blocked.

PR #1846, opened 2020-12-30
feat(origin-core-ui): improvements to user profile page

marcin-l-tsh:
#### before:
screencapture-localhost-3000-account-user-profile-2021-01-04-13_52_04
#### after:
screencapture-localhost-3000-account-user-profile-2021-01-04-13_42_48


packages/origin-ui-core/src/components/Account/UserProfile/ChangePassword.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+                            {isEditing && (
+                                <Button
+                                    style={{ marginRight: 10 }}
+                                    type="button"
+                                    variant="contained"
+                                    color="primary"
+                                    className="mt-3 right"
+                                    disabled={!(touched.currentPassword || touched.newPassword)}
+                                    onClick={async () => {
```
alexworker23, line 130: When clicking Save in ChangePassword the fields don't get empty. Save and Cancel buttons also don't hide.

packages/origin-ui-core/src/components/Account/UserProfile/Styles.tsx
```diff
@@ -0,0 +1,19 @@
...
+            display: 'flex',
+            alignItems: 'center'
+        },
+        infoIcon: {
+            marginLeft: '1rem'
+        }
+    })
+);
+
+export const useUserProfileStyle = () => useStyles(useTheme());
```
alexworker23: Why do you need UserProfile styles as a separate component? Why don't include it into the `UserDataEditForm.tsx` ?

packages/origin-ui-core/src/components/Account/UserProfile/UserDataEditForm.tsx
```diff
@@ -0,0 +1,260 @@
...
+                                </Grid>
+                            </Grid>
+                            {isEditing && (
+                                <Button
+                                    style={{ marginRight: 10 }}
+                                    type="button"
+                                    variant="contained"
+                                    color="primary"
+                                    className="mt-3 right"
+                                    disabled={
```
alexworker23, line 210: After first change of data Save button is no longer disabled.

packages/origin-ui-core/src/components/Account/UserProfile/UserDataEditForm.tsx
```diff
@@ -0,0 +1,260 @@
...
+                                    onClick={async () => {
+                                        await formikProps.validateForm();
+                                        await formikProps.submitForm();
+                                        setIsEditing(false);
+                                    }}
+                                >
+                                    {t('user.actions.save')}
+                                </Button>
+                            )}
+                            {isEditing && (
```
alexworker23, line 225: Same here. If you change some data and then click Cancel - your Save and Cancel buttons will no longer be disabled by default.

packages/origin-ui-core/src/components/Account/UserProfile/UserEmailChange.tsx
```diff
@@ -0,0 +1,153 @@
...
+                                    {t('user.actions.save')}
+                                </Button>
+                            )}
+                            {isEditing && (
+                                <Button
+                                    type="button"
+                                    variant="contained"
+                                    color="primary"
+                                    className="mt-3 right"
+                                    onClick={() => {
```
alexworker23, line 122: When changing email and then clicking Cancel - Save button is no longer disabled.

alexworker23 (changes requested):

josipbagaric:
@marcin-l-tsh Does something change on the user interface?
If so, could you attach screenshots to the PR? (Ideally a before/after comparison).

We do this for UI PRs to get a better sense of what's changed and how to review the PR.

packages/origin-ui-core/src/components/Account/UserProfile/Styles.tsx
marcin-l-tsh: it's also used in every other part of user profile page (ChangePassword, UserEmailChange, etc). should I use different approach is case like this?

alexworker23: Then you can create an scss file, like UserProfile.scss in styles folder and apply classes

packages/origin-ui-core/src/components/Account/UserProfile/ChangePassword.tsx
marcin-l-tsh: fixed, now after fields should be cleared after saving or canceling

packages/origin-ui-core/src/components/Account/UserProfile/UserDataEditForm.tsx
marcin-l-tsh: fixed
saving with success or canceling should now quit editing mod, then re-entering should keep save button disabled initially.

packages/origin-ui-core/src/components/Account/UserProfile/UserDataEditForm.tsx
marcin-l-tsh: fixed

packages/origin-ui-core/src/components/Account/UserProfile/UserEmailChange.tsx
marcin-l-tsh: fixed, now should be disabled after re-entering editing mode

marcin-l-tsh:
> @marcin-l-tsh Does something change on the user interface?
> If so, could you attach screenshots to the PR? (Ideally a before/after comparison).
> 
> We do this for UI PRs to get a better sense of what's changed and how to review the PR.

screenshots added

packages/origin-ui-core/src/components/Account/UserProfile/Styles.tsx
marcin-l-tsh: done, styles moved to scss file

alexworker23 (commented):
When user changes e-mail or password he is logged out, but the url stays the same `/account/user-profile`, which is causing the empty page to appear.
The solution is to configure redirect to the default page.

alexworker23 (approved):

PR #1859, opened 2021-01-05
fix(exchange-core-irec): show bids with no type filter when type filter specified

ioncreature:
Now bids list display bids without specified type even when market filter has.
@JonathanWfels take a look please. Is it ok?


kosecki123:
This change applies only to DeviceTypes, what about the other product elements that also use filtering?

ioncreature:
> This change applies only to DeviceTypes, what about the other product elements that also use filtering?

I've updated it to match region and operator

kosecki123 (approved):

PR #1860, opened 2021-01-05
feat(exchange): decouple accounting from main services

kosecki123:
This PR provides the decoupling of accounting for orders, transfers, and bundles from the main modules. This allows for more flexibility when implemented custom exchange implementation.


josipbagaric (approved):
👍

PR #1876, opened 2021-01-12
feat/ui lazy load

marcin-l-tsh:
In this PR: 
- Used Set for filtering out duplicate orders. 
Now Orders added to Redux state as an array instead of firing an event on each order one by one.

### DIFF
--- a/packages/exchange-ui-core/src/features/orders/reducer.ts
+++ b/packages/exchange-ui-core/src/features/orders/reducer.ts
@@ -17,8 +17,12 @@ export function ordersState<T>(
 ): IOrdersState {
     switch (type) {
         case OrdersActionsType.STORE_ORDERS:
-            const orders = [...state.orders.filter((o) => o.id !== payload.id)];
-            orders.push(payload);
+            const allOrders: Order[] = [...state.orders].concat(payload);
+            const set: Set<string> = new Set();
+            allOrders.map((order) => {
+                set.add(JSON.stringify(order));
+            });
+            const orders: Order[] = Array.from(set).map((order) => JSON.parse(order));
             return {
                 ...state,
                 orders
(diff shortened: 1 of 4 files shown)


packages/exchange-ui-core/src/features/orders/reducer.ts
```diff
@@ -1,21 +1,35 @@
...
+    isLoading: false
 };
 
 export function ordersState<T>(
     state: IOrdersState = initialState,
     { type, payload }: IOrderAction
 ): IOrdersState {
     switch (type) {
+        case OrdersActionsType.FETCH_ORDERS:
+            console.log('loading');
```
alexworker23, line 22: console.log remove

packages/exchange-ui-core/src/features/orders/reducer.ts
```diff
@@ -1,21 +1,35 @@
...
+            return {
+                ...state,
+                isLoading: true
+            };
+        case OrdersActionsType.FETCH_ORDERS_SUCCESS:
+            console.log('loaded');
```
alexworker23, line 28: console.log remove

packages/exchange-ui-core/src/features/bundles/reducer.ts
```diff
@@ -4,18 +4,30 @@ import { IBundleAction, BundlesActionType } from './actions';
...
+        case BundlesActionType.FETCH_BUNDLES:
+            return {
+                ...state,
+                isLoading: true
+            };
+        case BundlesActionType.FETCH_BUNDLES_SUCCESS:
+            return {
+                ...state,
+                isLoading: false
+            };
```
alexworker23, line 30: Why don't create a single action for setting the item loading state? 
Like setBundlesLoading(true/false)?

packages/exchange-ui-core/src/features/orders/selectors.ts
```diff
@@ -3,4 +3,6 @@ import { Order, Demand } from '../../utils/exchange';
 
 export const getOrders = (state: IExchangeState): Order[] => state.ordersState.orders;
 
+export const geIsLoading = (state: IExchangeState): boolean => state.ordersState.isLoading;
```
alexworker23, line 6: I guess this is a typo

alexworker23 (changes requested):
Outdated

PR #1881, opened 2021-01-15
chore(issuer-api): New api endpoint to get the SUM of certified energ…

arvindchez:
…y per deviceId


packages/issuer-api/src/pods/certificate/handlers/get-aggregare-certified-energy-by-device.handler.ts
```diff
@@ -0,0 +1,32 @@
...
+            deviceId,
+            fromTime: MoreThanOrEqual(startDate),
+            toTime: LessThanOrEqual(endDate)
+        });
+
+        if (!energies) {
+            return '0';
+        }
+
+        return `${energies.reduce((a, b) => a + (parseInt(b.energy, 10) || 0), 0)}`;
```
kosecki123: Please use BigNumber from `import { BigNumber } from 'ethers';` for the aritmetics, using js int can potentially overflow.

kosecki123 (approved):

PR #1904, opened 2021-01-19
feat(origin-backend-app): send email to organisation lead user email

ioncreature:
- [x] organization registration email
- [x] submitted generation device email 
- [x] certificate issuance approved email

I also added a lot of imports here and there


packages/origin-backend-app/src/notification/handlers/certificate-approved.handler.ts
```diff
@@ -0,0 +1,76 @@
...
+        if (this._deviceService) {
+            return this._deviceService;
+        }
+
+        this._deviceService = this.moduleRef.get<DeviceService>(DeviceService, {
+            strict: false
+        });
+
+        return this._deviceService;
+    }
```
ioncreature: @kosecki123 @JosephBagaric 
I do not like this, but I tried several other ways and they do not work. Maybe you have ideas how to inject `DeviceService`?

packages/origin-backend-app/src/origin-app.module.ts
```diff
@@ -79,10 +82,13 @@ export class OriginAppModule {
                 IssuerModule,
                 OrganizationModule,
                 UserModule,
-                CqrsModule
+                CqrsModule,
+                RegistrationModule
```
ioncreature, line 86: @kosecki123 @JosephBagaric 
This is crazy, `RegistrationModule` is part of `IRECOrganizationModule`, but if I remove this line the whole thing breaks:
```
Nest can't resolve dependencies of the CertificateApprovedHandler (MailService, UserService, ?, QueryBus, ModuleRef). Please make sure that the argument RegistrationService at index [2] is available in the OriginAppModule context.
@energyweb/origin-backend-app: Potential solutions:
@energyweb/origin-backend-app: - If RegistrationService is a provider, is it part of the current OriginAppModule?
@energyweb/origin-backend-app: - If RegistrationService is exported from a separate @Module, is that module imported within OriginAppModule?
```

packages/issuer-api/src/pods/certification-request/certification-request.module.ts
```diff
@@ -18,7 +19,7 @@ import { SyncCertificationRequestsTask } from './sync-certification-request.task
         CertificateModule
     ],
     controllers: [CertificationRequestController],
-    providers: [...Handlers, SyncCertificationRequestsTask],
-    exports: [...Handlers]
+    providers: [...Handlers, ...Events, SyncCertificationRequestsTask],
+    exports: [...Handlers, ...Events]
```
kosecki123: I don't think you need to have them as providers or exports. Event is regular POCO object, providers, exports are used for dependency injection purposes.

packages/issuer-api/src/pods/certification-request/handlers/approve-certification-request.handler.ts
```diff
@@ -63,16 +64,18 @@ export class ApproveCertificationRequestHandler
...
             approved: true,
             approvedDate: new Date(),
             issuedCertificateTokenId: newCertificateId
         });
 
+        this.eventBus.publish(
+            new CertificateCreatedEvent(newCertificateId, isPrivate ? { owner, energy } : null)
+        );
+
+        this.eventBus.publish(new CertificateRequestApprovedEvent(id));
```
kosecki123, line 77: What's the purpose of the `CertificateRequestApprovedEvent` ?

kosecki123 (dismissed):

packages/issuer-api/src/pods/certification-request/handlers/approve-certification-request.handler.ts
ioncreature: CertificateCreated event fired too often and there is no way I can check if I already sent an email. So instead of I decided to create a separate event that I can fully control.

packages/origin-backend-app/src/origin-app.module.ts
kosecki123: Works as expected: "The module encapsulates providers by default. This means that it's impossible to inject providers that are neither directly part of the current module nor exported from the imported modules. Thus, you may consider the exported providers from a module as the module's public interface, or API"

kosecki123: In this example, `IRECOrganizationModule` is probably overhead because it just import a single module.

packages/issuer-api/src/pods/certification-request/certification-request.module.ts
ioncreature: Fixed, thanks

packages/origin-backend-app/src/notification/handlers/certificate-approved.handler.ts
kosecki123: The problem is that the DeviceModule is currently a dynamic module that messes up with the dependency injection. For now I do know any better solution.

packages/issuer-api/src/pods/certification-request/handlers/approve-certification-request.handler.ts
kosecki123: @JosephBagaric please take a look

josipbagaric (approved):

PR #1923, opened 2021-01-25
feat(exchange-ui-core): use set instead of loop in redux

alexworker23:
Used Set for filtering out duplicate orders. 
Now Orders added to Redux state as an array instead of firing an event on each order one by one. 
This adjustment was made due to poor performance of the previous order store strategy.

### DIFF
--- a/packages/exchange-ui-core/src/features/orders/reducer.ts
+++ b/packages/exchange-ui-core/src/features/orders/reducer.ts
@@ -17,8 +17,12 @@ export function ordersState<T>(
 ): IOrdersState {
     switch (type) {
         case OrdersActionsType.STORE_ORDERS:
-            const orders = [...state.orders.filter((o) => o.id !== payload.id)];
-            orders.push(payload);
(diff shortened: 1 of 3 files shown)


packages/exchange-ui-core/src/features/orders/reducer.ts
```diff
@@ -17,8 +17,12 @@ export function ordersState<T>(
 ): IOrdersState {
     switch (type) {
         case OrdersActionsType.STORE_ORDERS:
-            const orders = [...state.orders.filter((o) => o.id !== payload.id)];
-            orders.push(payload);
+            const allOrders: Order[] = [...state.orders].concat(payload);
+            const set: Set<string> = new Set();
+            allOrders.map((order) => {
```
josipbagaric: Are you sure you should be using `.map()` here and not `.forEach()`?

josipbagaric (approved):

ioncreature (approved):

PR #1926, opened 2021-01-27
feat/demand date filter

alexworker23:
This PR adds to table filter a new functionality: **Dependant Filters**.
Using this feature user can change one filter look if this filters dependency has changed. 
Dependency rules are passed from the top level table component.
Exact implementation is connected with Demands Table in My Orders.
Case: 
User wants to filter the demands table. He chooses the period of Day/Week
Here is his Start/End date inputs will look like:
day-filter
So he will be able to choose day precisely.
If he chooses Year/Month his inputs look like this:
month-filter


josipbagaric (approved):
👍

PR #1928, opened 2021-01-27
chore(exchange-io-erc1888): Move supply watching functionality from e…

arvindchez:
…rc1888 watcher exchange using eventbus


kosecki123:
Based on the changes this PR does 2 things:
- declares events in the `exchange` package
- moved code to handlers in the `exchange-io-erc1888`
but this does not achieve the goal of the PR which is to move the "logic" of post for sale from `exchange-io-erc1888` to the `exchange` package. Note that post for sale is not dependent on the method of "discovering" the deposit thus can be reused in case of other `exchange-io-*` packages.

kosecki123 (changes requested):
As commented in the PR.

josipbagaric (commented):
So in order for this to work we need to:
1.  Move the event handlers to the `@energyweb/exchange` packages
2. Move `this.tryPostForSale()` functionality from `@energyweb/exchange-io-erc1888` to `@energyweb/exchange` 

The deposit watcher should only trigger the `DepositDiscoveredEvent` event and the rest should be handled by the exchange.

Did I get this correctly @kosecki123?

packages/exchange/src/pods/post-for-sale/handlers/deposit-discovered-event.handler.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    public async handle(event: DepositDiscoveredEvent): Promise<void> {
+        const { address, amount, asset } = event.deposit;
+        this.logger.debug(
+            `Deposit discovered event raised for deposit with tokenId=${asset.tokenId} from=${address} with value=${amount}`
+        );
+
+        this.transferService.storeDeposit(event.deposit);
+    }
+}
```
kosecki123: This file should be in `transfer` pod. Storing deposit has nothing to do with posting for sale.

packages/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
```diff
@@ -2,22 +2,20 @@ import { Contracts } from '@energyweb/issuer';
...
 
 import {
     IExchangeConfigurationService,
-    IExternalDeviceService,
-    AccountService,
-    CreateAskDTO,
-    OrderService,
-    TransferStatus,
-    TransferService
+    TransferService,
```
kosecki123, line 12: This can be removed I guess.

kosecki123 (changes requested):
Looks good 👍 2 changes request in the comments.

packages/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
arvindchez: The transfer service is needed for below code block. i.e to get the confirmation block. Please confirm.    

 const blockNumber = await this.transferService.getLastConfirmationBlock();

        this.logger.debug(`Starting from block ${blockNumber}`);

        this.provider.resetEventsBlock(blockNumber);
        this.provider.on(
            {
                address: this.registryAddress,
                topics
            },
            (event: providers.Log) => this.processEvent(event)
        );

josipbagaric (changes requested):
The PR doesn't follow our commit scheme. Please rebase the PR and make sure that all commit descriptions match our commit scheme.

arvindchez:
Send out a new clean PR (https://github.com/energywebfoundation/origin/pull/1932) with all the changes as per comments here.

PR #1931, opened 2021-01-28
chore(exchange-io-erc1888): Move supply watching functionality from e…

arvindchez:
…rc1888 watcher to exchange using event modelling


josipbagaric:
The commit scheme is still incorrect.

Commit `chore(exchange-io-erc1888): ...` does changes to both the `exchange-io-erc1888` and `exchange` packages. This commit description is misleading. This messes up our releases because automatic releases will only bump the version of the `exchange-io-erc1888` package.

Please do **one** of the following:
1. _**(Less verbose but works)**_ Don't specify the package. Rename the commit to `chore: <description>`. This will bump the patch version of the semver to all the packages that have been changed.
2. _**(More verbose, better IMO)**_ Create 2 commits, one for exchange-io-erc1888 called `chore(exchange-io-erc1888): <description>` and another one for exchange called `chore(exchange): <description>` with the respective changes to both packages.

PR #1934, opened 2021-01-29
chore: User code cleanup

josipbagaric:
- Add `BlockchainAccountGuard` and `BlockchainAccountDecorator` to issuer-api
- Streamline User DTOs
- Remove unused user service functions
- Remove unused UI user functions and calls


packages/origin-ui-core/src/components/admin/AdminUserForm.tsx
```diff
@@ -77,11 +77,9 @@ export function AdminUserForm(props: IProps) {
         dispatch(setLoading(true));
 
         try {
-            const formData: UserUpdateData = {
-                ...values
-            };
-
-            await adminClient.updateUser(values.id.toString(), formData);
+            await adminClient.updateUser(values.id, {
```
alexworker23: Admin is now prohibited to change User data ?

kosecki123 (approved):

packages/origin-ui-core/src/components/admin/AdminUserForm.tsx
josipbagaric: This is a good point. Somehow I missed this because UpdateUserDTO only contained the notifications and blockchain account, but we forwarded the whole user object and it accidentally worked, but it shouldn't have:
https://github.com/energywebfoundation/origin/pull/1934/files#diff-f45e6838c1a09e962293d393c9c0b7260b0bfdfbbbcd2600670479017c71c000R4

Will retest this and make sure it's working as expected.

josipbagaric: Done. Fixed it!

alexworker23 (approved):

PR #1946, opened 2021-02-01
fix(exchange-io-erc1888): deposit watcher amount to string

josipbagaric:
Fixes an error when depositing.

```
[TransferService] invalid input syntax for type bigint: "{"type":"BigNumber","hex":"0x3b9aca00"}" +12ms
```

Additionally:
- Enables `exchange-io-erc1888` tests
- Fixes an error where `post-for-sale` module was not even imported into the main app

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
         "test:utils-general": "lerna run test --scope @energyweb/utils-general --stream",
         "test:ci:contracts": "lerna run --scope @energyweb/issuer --scope @energyweb/utils-general test:concurrent --since master --parallel --stream",
         "test:ci:apps": "lerna run --scope @energyweb/exchange --scope @energyweb/origin-backend test:concurrent --since master --parallel --stream",
-        "test:ci:e2e": "lerna run test:e2e --stream --ignore @energyweb/exchange-* && lerna run test:e2e --stream --scope @energyweb/exchange-irec",
+        "test:ci:e2e": "lerna run test:e2e --stream --ignore @energyweb/exchange-irec && lerna run test:e2e --stream --scope @energyweb/exchange-irec",
         "publish:canary": "lerna publish --yes --skip-git --exact --cd-version=prerelease --pre-dist-tag canary --preid=alpha.$TRAVIS_BUILD_NUMBER",
         "publish:release": "lerna version --create-release github --conventional-commits --exact --yes --message \"chore(release): publish /skip-deploy\" && lerna publish from-git --yes",
         "build:containers:canary": "lerna run build:container:canary --stream",
--- a/packages/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
+++ b/packages/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
@@ -92,6 +92,12 @@ export class DepositWatcherService implements OnModuleInit {
 
         const { _from: from, _to: to, _value: value, _id: id } = log;
 
+        if (!value || !from || !to) {
+            this.logger.error(`Received an incorrect even: ${JSON.stringify(event)}`);
+
+            return;
+        }
+
         if (to !== this.walletAddress) {
             this.logger.debug(
                 `This transfer is to other address ${to} than wallet address ${this.walletAddress}`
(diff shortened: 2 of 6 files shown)


packages/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
```diff
@@ -107,8 +107,8 @@ export class DepositWatcherService implements OnModuleInit {
 
         const deposit: CreateDepositDTO = {
             transactionHash,
-            address: from as string,
-            amount: value as string,
+            address: from,
+            amount: value.toString(),
```
ioncreature, line 117: It is safer to use `String(value)` (If `value` can be `null` or `undefined`)

ioncreature (dismissed):

kosecki123:
Where are the integration tests? Such a basic error should be caught already.

packages/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
josipbagaric: Fair point, but the ethers lib always returns a `value`, of type ethers BigNumber. `.toString()` is an overriden function of BigNumber, not the generic JS one.

Besides, I don't think `String(value)` would work properly for null or undefined because the event expects a stringed number, not just any string.

Screen Shot 2021-02-01 at 15 32 06

josipbagaric: Good point though, we should add a check to not emit the event if the value is incorrect.

josipbagaric: Added.

josipbagaric:
> Where are the integration tests? Such a basic error should be caught already.

Added!

kosecki123 (approved):

package.json
```diff
@@ -62,7 +62,7 @@
         "test:utils-general": "lerna run test --scope @energyweb/utils-general --stream",
         "test:ci:contracts": "lerna run --scope @energyweb/issuer --scope @energyweb/utils-general test:concurrent --since master --parallel --stream",
         "test:ci:apps": "lerna run --scope @energyweb/exchange --scope @energyweb/origin-backend test:concurrent --since master --parallel --stream",
-        "test:ci:e2e": "lerna run test:e2e --stream --ignore @energyweb/exchange-* && lerna run test:e2e --stream --scope @energyweb/exchange-irec",
+        "test:ci:e2e": "lerna run test:e2e --stream --ignore @energyweb/exchange-irec && lerna run test:e2e --stream --scope @energyweb/exchange-irec",
```
ioncreature, line 65: I think we have to return `*` here

packages/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
```diff
@@ -92,6 +92,12 @@ export class DepositWatcherService implements OnModuleInit {
 
         const { _from: from, _to: to, _value: value, _id: id } = log;
 
+        if (!value || !from || !to) {
+            this.logger.error(`Received an incorrect even: ${JSON.stringify(event)}`);
```
ioncreature: Typo `even` => `event`

packages/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
ioncreature: under the hood `String()` calls `.toString()` method if possible, so it is fine to use it all the time.

ioncreature: But here `.toString()` is ok since we have validation above

package.json
josipbagaric: Why? I removed this to enable `exchange-io-erc1888` tests.

packages/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
josipbagaric: I would disagree on using `String()`. I would prefer to get an error when `value == undefined` than to convert `undefined` to `'undefined'`. When it throws an error it's simpler to debug than getting an unexpected string.

package.json
ioncreature: Ok, I see. then it is fine

ioncreature (approved):

PR #1947, opened 2021-02-01
feat(issuer-irec-api-wrapper): irec device management api

ioncreature:
Added methods to call IREC Evident API
1. Organization APIs
2. Issue request API
3. Device API
4. Fuel API

Also added test to check device and issue request flows


kosecki123 (approved):
👍

PR #1948, opened 2021-02-02
Getting certified energy amount without auth

kosecki123:
This PR fixes the issue where certified energy was calculated for device owner only, also since now, the data can be queried by the non-authenticated users, which is used by the public device details page.


alexworker23 (approved):

josipbagaric (approved):
👍

PR #1950, opened 2021-02-02
chore: distinguish between device services

josipbagaric:
We have a DeviceService defined in both `@energyweb/origin-backend` and `@energyweb/origin-device-registry-irec-local-api-client`

Getting `DeviceService` from NestJS's moduleRef would sometimes returns the incorrect DeviceService due to a naming collision.

Solution:
- Distinguish the two existing `DeviceService` using interfaces and symbols


kosecki123:
This has been fixed by #1987

PR #1957, opened 2021-02-04
Separate packages for legacy I-REC device management

kosecki123:
This PR makes the legacy "device" registration functionality as a separate package. Additionally, tests fixes, log suppression when running the tests.


packages/origin-backend/src/pods/user/dto/user.dto.ts
```diff
@@ -32,7 +33,7 @@ export class UserDTO implements IUser {
     lastName: string;
 
     @ApiProperty({ type: String })
-    @IsNotEmpty()
+    @IsEmail()
     @IsString()
```
josipbagaric, line 37: Is removing `@IsNotEmpty()` intentional? Or did you want to replace `@IsString` with `@IsEmail`?

josipbagaric (approved):

packages/origin-backend/src/pods/user/dto/user.dto.ts
kosecki123: In fact, having `IsEmail` should be enough.

josipbagaric: We also don't need IsString if we have IsEmail

alexworker23 (approved):

PR #1990, opened 2021-02-09
chore: supply to exchange ui

alexworker23:
In this PR 
Supply is moved from `Device` to `Exchange`.
Created a separate Redux state and saga for supply. 
Fixed Supply filters


packages/ui/exchange-ui-core/src/containers/SupplyTable.tsx
```diff
@@ -0,0 +1,229 @@
...
+} from '@energyweb/origin-ui-core';
+import { UpdateSupplyModal, RemoveSupplyConfirmation } from '../components/modal';
+import { getEnvironment } from '../features/general';
+import { IDeviceWithSupply, ISupplyTableRecord } from '../types';
+import { getSupplies, fetchSupplies } from '../features/supply';
+
+export const KeyStatus = {
+    Active: 'Active',
+    Paused: 'Paused'
+};
```
josipbagaric: Probably better suited to be an `enum`. Also maybe better to rename it to `SupplyStatus` to make the intention clearer.

```
export enum SupplyStatus {
    Active = 'Active',
    Paused = 'Paused'
}
```

packages/ui/exchange-ui-core/src/containers/SupplyTable.tsx
```diff
@@ -0,0 +1,229 @@
...
+                type: CustomFilterInputType.string
+            }
+        },
+        {
+            property: (device: IDeviceWithSupply) =>
+                device.active ? KeyStatus.Active : KeyStatus.Paused,
+            label: t('exchange.supply.status'),
+            input: {
+                type: CustomFilterInputType.dropdown,
+                availableOptions: STATUS_OPTIONS
```
josipbagaric: If you make `KeyStatus` an enum, you can remove `STATUS_OPTIONS` and only put `availableOptions: Object.values(KeyStatus)`

josipbagaric (approved):

packages/ui/exchange-ui-core/src/types/supply.ts
```diff
@@ -0,0 +1,20 @@
+import { SupplyDto, UpdateSupplyDto } from '@energyweb/exchange-client';
+import { IDevice } from '@energyweb/origin-backend-core';
+
+export interface IDeviceWithSupply extends IDevice, Partial<Pick<SupplyDto, 'active' | 'price'>> {
```
kosecki123: I think I would not make any binding between IDeviceWithSupply and IDevice, just define the fields you need here, like name and others. This will be useful when we would push data from `local` and other device registries. IDevice is considered old interface that is only compatible with `...device...form` api.

kosecki123 (approved):

PR #1991, opened 2021-02-09
fix: skip dockerhub publishing due to reliability issues

kosecki123:
This PR removes the step of publishing artifacts as Dockerhub images. The main reason is that recently Dockerhub stability is very spotty, causing random CI build errors.


josipbagaric:
Does this mean we will have to do manual deployments?

kosecki123:
> Does this mean we will have to do manual deployments?

No, it means that we skip pushing to dockerhub, pushing to heroku registry stays intact.

josipbagaric (approved):

josipbagaric:
Cool. I thought Heroku pulls the images from DockerHub. Approved 👍

PR #1999, opened 2021-02-11
feat: add endpoints to get device and fuel types

ioncreature:


packages/devices/origin-device-registry-irec-local-api/src/device/device.controller.ts
```diff
@@ -52,10 +57,36 @@ export class DeviceController {
...
+        description: 'Returns all IREC fuels types'
+    })
+    getFuelTypes(): CodeNameDTO[] {
+        const fuelTypes = this.deviceService.getFuelTypes();
+
+        return fuelTypes.map((fuelType) => plainToClass(CodeNameDTO, fuelType));
     }
 
-    @Get('/:id')
+    @Get('/device/:id')
```
ioncreature, line 89: I've added "device" prefix here and for @Put('/device/:id') because otherwise it matches to GET /fuel-type and /device-type

ioncreature: @kosecki123

kosecki123 (approved):

PR #2004, opened 2021-02-11
fix(origin-backend-client): fix file upload regression

josipbagaric:


kosecki123 (approved):
I don't get it but LGTM I guess 😃

alexworker23 (approved):

josipbagaric:
> I don't get it but LGTM I guess 😃

This file does simple string replacement to replace autogenerated code and fix uploads. (kind of a hack to make it work)

Since the dependency update changed the string that needs to be replaced, no string was replaced, and broke the uploads.

PR #2006, opened 2021-02-11
feat: irec device and fuel types

ioncreature:
- [x] Device and fuel type validation
- [x] Create device-fuel map(this waits for the answer from IREC on how to implement it)


packages/tools/migrations-irec/config/demo-config.json
```diff
@@ -89,64 +89,31 @@
...
+        ["ES420", "TC421"],
+        ["ES420", "TC422"],
+        ["ES420", "TC423"],
+        ["ES420", "TC424"],
+
+        ["ES430", "TC431"],
+        ["ES430", "TC432"],
+
+        ["ES440", "TC441"],
+        ["ES440", "TC442"],
```
ioncreature: Since IREC does not have any mapping between fuel and device codes, I decided to put here just a small subset of available types. Otherwise, I would have to put here a multiplication of 38 device types and 31 fuel types.

Also Dan said that they are going to implement a new system behind it, so we will have to update this list soon.

packages/apps/origin-backend-irec-app/src/integration/external-device.service.ts
```diff
@@ -25,9 +25,15 @@ export class ExternalDeviceService implements IExternalDeviceService {
...
 
-        // TODO: proper mapping
+        const [deviceTypeCode, fuelTypeCode] = device.deviceType.split(';');
+        const deviceTypes = this.deviceService.getDeviceTypes();
+        const fuelTypes = this.deviceService.getFuelTypes();
+
+        const deviceType = deviceTypes.find((value) => value.code === deviceTypeCode)?.name ?? '';
+        const fuelType = fuelTypes.find((value) => value.code === fuelTypeCode)?.name ?? '';
+
         return {
```
kosecki123, line 34: Instead of defaulting to empty string: which would eventually end up with `;` I suggest we throw the exception and make sure ask is not created.

packages/tools/migrations-irec/config/demo-config.json
kosecki123: Use `name` instead of `code`.

kosecki123 (changes requested):
Comments ☝️

packages/apps/origin-backend-irec-app/src/integration/external-device.service.ts
ioncreature: done

kosecki123 (approved):

PR #2035, opened 2021-02-15
Feat/blockchain inbox

marcin-l-tsh:
image
image


packages/ui/origin-ui-core/src/components/certificates/certificateMenuCreator.tsx
```diff
@@ -64,6 +64,13 @@ function ApprovedCertificationRequestsTable() {
     return <CertificationRequestsTable approved={true} />;
 }
 
+function EnchangeInboxPage(): JSX.Element {
+    return <ExchangeInbox mode={CertificateSource.Exchange} />;
+}
+function BlockchainInboxPage(): JSX.Element {
+    return <ExchangeInbox mode={CertificateSource.Blockchain} />;
```
alexworker23: If you are using this component as root for several other components you should have an abstract naming to avoid confusion in future development process.
`CertificatesInbox` or `Inbox` for example.

packages/ui/origin-ui-core/src/components/certificates/ExchangeInbox.tsx
```diff
@@ -320,7 +343,35 @@ export function ExchangeInbox(): JSX.Element {
         });
     }
 
-    const SelectedCertificatesList = (): [IInboxItemData, IInboxCertificateData][] => {
+    async function deposit() {
+        const certs = getSelectedCertificates();
+        certs.forEach((cert) => {
+            dispatch(
```
alexworker23: If you want to make this component an abstract and reusable - such functions should be passed here as props.

packages/ui/origin-ui-core/src/components/certificates/ExchangeInbox.tsx
```diff
@@ -336,11 +387,6 @@ export function ExchangeInbox(): JSX.Element {
...
     return (
         <div>
-            <InboxItemPreview
-                open={modalData !== null}
-                setOpen={() => setModalData(null)}
-                data={modalData}
-            />
             <Grid container spacing={3}>
                 <Grid item xs={12} md={7}>
                     <div className={classes.box}>
```
alexworker23: Title for the inbox should be passed as prop

packages/ui/origin-ui-core/src/components/certificates/ExchangeInbox.tsx
```diff
@@ -320,7 +343,35 @@ export function ExchangeInbox(): JSX.Element {
...
+                })
+            );
+        });
+    }
+
+    async function claim() {
+        const certs = getSelectedCertificates();
+
+        dispatch(
+            requestClaimCertificateBulk({
```
alexworker23: `claim` and `deposit` functions are not working, please check the reasons for it

packages/ui/origin-ui-core/src/components/certificates/certificateMenuCreator.tsx
```diff
@@ -64,6 +64,13 @@ function ApprovedCertificationRequestsTable() {
     return <CertificationRequestsTable approved={true} />;
 }
 
+function EnchangeInboxPage(): JSX.Element {
+    return <ExchangeInbox mode={CertificateSource.Exchange} />;
+}
+function BlockchainInboxPage(): JSX.Element {
```
alexworker23: Maybe it is better to remove these 2 components from `menuCreator` and make them separate components. 
Including Table specific functions like `claim`, `deposit` for each of them and passing these functions as props to general inbox component.
By doing this you will make Inbox general component more lightweight, easy to read and more reusable.

alexworker23 (changes requested):

alexworker23:
If the Blockchain Inbox item has 0 energy - it should not be displayed
blockchain-inbox

alexworker23:
Also I think a good edition to improve user experience would be:
Submit buttons like `Sell`/`Withdraw`/`Deposit` should be disable when the user enters the Editing mode of the current Selected item

alexworker23 (approved):

PR #2036, opened 2021-02-16
chore/energy api for smart meter graphs

alexworker23:
In this PR:
Created a separate UI graph for Smart Meter Readings, which is connected to energy-api


packages/ui/origin-ui-core/src/features/general/actions.ts
```diff
@@ -31,6 +31,7 @@ export interface IEnvironment {
     GOOGLE_MAPS_API_KEY: string;
     MARKET_UTC_OFFSET: number;
     DISABLED_UI_FEATURES: string;
+    SMART_METER_ID: string;
```
josipbagaric, line 34: Where is this used and what for?

josipbagaric (approved):

packages/ui/origin-ui-core/src/features/general/actions.ts
alexworker23: Will be used for extracting `meterId` from device's `externalDeviceIds` array in the same way we are using `ISSUER_ID` now

PR #2040, opened 2021-02-18
fix/i18next alias

alexworker23:
In this PR removed i18nextProviders from UI packages adapters. Instead used a webpack alias for react-i18next


josipbagaric (approved):
👌

marcin-l-tsh (approved):

PR #2041, opened 2021-02-19
DOI-688 use github actions

GrzegorzMatuszakTSH:


kosecki123 (approved):
Looks good ! 👍 

One missing thing here is to disable Travis, I guess removing / renaming the travis.yml would be ok for now.

PR #2042, opened 2021-02-19
feat(origin-organization-irec-local-api): irec device creation, refresh irec tokens

ioncreature:
- [x] Create device on IREC side
- [x] Refresh access tokens
- [x] Use IREC device statuses
- [x] Edit device 
- [x] Submit device to IREC
- [x] Check device state cron


kosecki123 (changes requested):
So DeviceService requires now working I-REC API connection, and there is no way to test our UI without it. Ideally, your changes should go to a separate device service, like `device-irec.service.ts` and previous functionality would stay. Same with cron jobs, they only should work when we are using I-REC api connection.

kosecki123:
> So DeviceService requires now working I-REC API connection, and there is no way to test our UI without it. Ideally, your changes should go to a separate device service, like `device-irec.service.ts` and previous functionality would stay. Same with cron jobs, they only should work when we are using I-REC api connection.

Additionally, since the module is imported to `origin-backend-irec-app`, it would be great if the package exposes 2 separate modules, one with I-REC enabled code (service, cron) and one without.

kosecki123:
Handlebars import issue has been fixed in #2105

PR #2066, opened 2021-02-22
Feat/device icons svg change

marcin-l-tsh:
updated svg Icons
image

### DIFF
--- a/packages/ui/origin-ui-core/src/components/Icons/DeviceIcon.tsx
+++ b/packages/ui/origin-ui-core/src/components/Icons/DeviceIcon.tsx
@@ -1,40 +1,44 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
-import solar from '../../../assets/icon_solar.svg';
-import wind from '../../../assets/icon_wind.svg';
-import hydro from '../../../assets/icon_hydro.svg';
-import iconThermal from '../../../assets/icon_thermal.svg';
-import iconSolid from '../../../assets/icon_solid.svg';
-import iconLiquid from '../../../assets/icon_liquid.svg';
-import iconGaseous from '../../../assets/icon_gaseous.svg';
-import iconMarine from '../../../assets/icon_marine.svg';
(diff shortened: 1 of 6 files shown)


packages/ui/origin-ui-core/src/components/certificates/Inbox/InboxItem.tsx
```diff
@@ -157,7 +160,12 @@ export function InboxItem(props: {
                                     onChange={() => onCertificateSelect(cert.id, device.id)}
                                 />
                             </div>
-                            <div className={classes.iconContainer}>
+                            <div
+                                className={classes.iconContainer}
+                                style={{
+                                    fill: isSelected ? PRIMARY_COLOR : '#969696'
```
alexworker23: `#969696` could be replaced using `LightenColor()`

packages/ui/origin-ui-core/src/components/Icons/DeviceIcon.tsx
```diff
@@ -1,40 +1,44 @@
...
+import { ReactComponent as IconSolar } from '../../../assets/device/icon-solar.svg';
 
 export function DeviceIcon(props: { type: string; className: string }): JSX.Element {
     const { t } = useTranslation();
     const { type, className } = props;
-    let image = solar;
+    let SvgImage = <IconSolar />;
     let tooltip = '';
 
     if (type.startsWith('Wind')) {
```
alexworker23: You can move this condition into a function, and call it inside the component return statement removing variables assigned with `let`. 
You will have 2 separate functions: 1 for assigning tooltip and 1 for choosing matching Icon.

packages/ui/origin-ui-core/src/components/certificates/Inbox/InboxItem.tsx
marcin-l-tsh: done

alexworker23 (approved):

PR #2080, opened 2021-02-25
chore/refactor device UI and add device data layer

alexworker23:
In this PR:
1. Normalize Devices and Certificates data flow. 
Which means refactoring those 2 parts of UI according to selected Data Fetching strategy. 
Cleaned up the `origin-ui-core` redux files.
2. Added manageable data layer for `exchange-ui-core`, which is now selected from root `origin-ui` package. This allows `exchange-ui-core` to switch between 2 different device API's
3. For `origin-irec-ui-core` added Certificates - as a temporary solution for managing 2 different device API's inside Certificates.
4. For `origin-device-registry-api` and `origin-device-irec-local-api` adjusted DTOs, so that they will appear in auto-generated client.


kosecki123 (approved):
🔥

PR #2112, opened 2021-03-04
feat(origin-organization-irec-local-api): irec device creation, refresh irec tokens

ioncreature:
- [x] Create device on IREC side
- [x] Refresh access tokens
- [x] Use IREC device statuses
- [x] Edit device 
- [x] Submit device to IREC
- [x] Check device state cron 
- [x] Allow package to work without irec integration


packages/devices/origin-device-registry-irec-local-api/src/device/device.service.ts
```diff
@@ -49,20 +62,23 @@ export class DeviceService {
...
         const device = await this.findOne(id);
 
         if (!device) {
             return null;
         }
 
-        await this.repository.update(device.id, { status });
-
-        this.eventBus.publish(new DeviceStatusChangedEvent(device, status));
+        const irecDevice = await this.irecDeviceService.update(user, device.code, deviceData);
```
kosecki123, line 72: According to I-REC API documentation, editing is not allowed when `Updates can be made to devices with any status except withdrawn and approved.` I think we should handle the error and give the appropriate error, otherwise the users will get error 500.

packages/traceability/issuer-irec-api-wrapper/src/IRECAPIClient.ts
```diff
@@ -457,6 +464,7 @@ export class IRECAPIClient {
             response.data.refresh_token,
             response.data.expires_in
         );
+        this.emit('tokensRefreshed', this.accessTokens);
```
kosecki123, line 467: Not sure why we need to use the mechanism of event emitting here? Could you explain your thought process here?

ioncreature: Since we are storing all the access tokens we have to update them in DB every time it refreshes. So events are the obvious approach to use.
If you have a better idea, I'm ready to listen.

packages/devices/origin-device-registry-irec-local-api/src/device/device.service.ts
ioncreature: Thanks for noticing it. Actually, it allowed editing the device in all states except "InProgress"

kosecki123 (approved):

PR #2136, opened 2021-03-08
feat: clear matching engine cache

josipbagaric:
Adds an option to reset the in-memory storage of the matching engine.


packages/trade/exchange/src/pods/matching-engine/matching-engine.service.ts
```diff
@@ -99,6 +89,12 @@ export class MatchingEngineService<TProduct, TProductFilter> implements OnModule
         this.matchingEngine.cancelOrder(orderId);
     }
 
+    public async clear(): Promise<void> {
+        this.matchingEngine.clear();
```
kosecki123: Maybe set initialized flag to false so the tick won't get in a way when clearing and reloading, then set to true after all operation completed.

kosecki123 (approved):

PR #2137, opened 2021-03-09
Feat/admin claims table

marcin-l-tsh:
Added page with table of claims for Admin
image

### DIFF
--- a/packages/ui/origin-ui-core/src/components/admin/AdminClaimsTable.tsx
+++ b/packages/ui/origin-ui-core/src/components/admin/AdminClaimsTable.tsx
@@ -1,0 +1,74 @@
+import React, { useEffect, useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import {
+    fetchAllDevices,
+    getAllDevices,
+    getBackendClient,
+    getCertificatesClient,
+    getEnvironment
+} from '../../features';
+import { TableMaterial } from '../Table';
+
+interface IRow {
+    id: number;
+    device: string;
+    energy: string;
+    beneficiary: string;
+}
+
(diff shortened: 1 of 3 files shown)


packages/ui/origin-ui-core/src/components/admin/AdminClaimsTable.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        beneficiary: certDevice.organizationName
+                    } as IRow;
+                })
+            );
+        }
+    }
+
+    useEffect(() => {
+        update();
+    }, [allDevices]);
```
alexworker23, line 56: Add false check here

packages/ui/origin-ui-core/src/components/admin/AdminClaimsTable.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    async function update() {
+        const { data: certs } = await certificatesClient.getAll();
+
+        if (certs) {
+            setRows(
+                certs.map((c) => {
+                    const certDevice = allDevices.find((device) =>
+                        device.externalDeviceIds.some(
+                            (deviceExternalId) =>
```
alexworker23, line 37: Please, avoid doing such long calculation inside `setState`. 
Shape the data first, assign it to a variable and pass as argument to setState function.

alexworker23 (changes requested):
As far as I can understand you are now displaying the wrong data from the certificate. 
You should be working with `claims` array, right @kosecki123?
image

kosecki123:
> As far as I can understand you are now displaying the wrong data from the certificate.
> You should be working with `claims` array, right @kosecki123?
> image

Yes, please use `claims` array of the certificate object.

PR #2140, opened 2021-03-10
Added claiming period to on-chain claims

kosecki123:
This PR adds fromDate and toDate to claiming on-chain payload.


packages/traceability/issuer-api/src/pods/certificate/commands/claim-data.dto.ts
```diff
@@ -0,0 +1,34 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsDateString, IsOptional } from 'class-validator';
+
+export class ClaimDataDTO {
+    @ApiProperty({ type: String, required: false })
+    @IsOptional()
+    beneficiary?: string;
```
josipbagaric, line 8: You might want to add `@IsString()` validation to these string fields.

josipbagaric (approved):

PR #2144, opened 2021-03-11
feat/admin claims

alexworker23:
1. Now Admin can see all the claims done by the users.
image
***
2. Added additional fields for claiming: fromDate & toDate. 
If the user doesn't change them, by default it takes the current date for both fields.

- New input fields(Datepickers) added to Beneficiary form inside new Blockchain Inbox:
image
- New input fields(Datepickers) added to Claim Modal inside old Certificates Inbox:
image
- Added time info into Certificate Detail View
image

### DIFF
--- a/packages/ui/origin-ui-core/src/features/certificates/types.ts
+++ b/packages/ui/origin-ui-core/src/features/certificates/types.ts
@@ -1,3 +1,4 @@
+import { ClaimDTO } from '@energyweb/issuer-api-client';
 import { BigNumber } from 'ethers';
 
 export interface IEnergy {
@@ -15,7 +16,7 @@ export interface ICertificate {
     energy: IEnergy;
     isOwned: boolean;
     isClaimed: boolean;
-    myClaims?: any[];
+    myClaims?: ClaimDTO[];
 
     // Blockchain specific properties
     blockchain?: any;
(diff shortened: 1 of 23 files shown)


packages/ui/origin-ui-core/src/features/certificates/types.ts
```diff
@@ -15,7 +16,7 @@ export interface ICertificate {
     energy: IEnergy;
     isOwned: boolean;
     isClaimed: boolean;
-    myClaims?: any[];
+    myClaims?: ClaimDTO[];
```
josipbagaric, line 19: I don't think we need to have this type `ICertificate` defined in the UI packages. 

It's better to use the `CertificateDTO` class from `@energyweb/issuer-api-client`, so that we don't have to update these UI classes every time, but they are automatically updated through the client.

josipbagaric (approved):
LGTM, minor comments.

Also, shouldn't we show the energy in the format `10 MWh` instead of `10MWh` like in the screenshots?

packages/ui/origin-ui-core/src/features/certificates/types.ts
josipbagaric: We used to define those classes manually because we didn't have the automatically generated definitions from the client, and we didn't want to import the backend packages into the UI package.

So I consider these "copied" types to be legacy.

alexworker23: I totally agree with you, but I suggest making this switch as a separate PR, because there are a lot of places where we should change it e.g. exchange-ui-core

PR #2191, opened 2021-03-18
feat(origin-ui-irec-core): device import ui

ioncreature:


alexworker23:
As far as I understand this still could not be fully implemented? 
Because there is a mock on Device Import UI and navigating to page leads to 403 error for missing connection.

alexworker23 (approved):

PR #2192, opened 2021-03-19
fix/getting meter reads after vital device data

alexworker23:
When getting All/My devices - at first device data is received without meterReads, allowing to display table data. 
At the same moment second request is started at the background, where devices with meter reads are received. 
After this we are merging devices with meterRead and without and update the stored data.

### DIFF
--- a/packages/ui/origin-ui-core/src/components/devices/Table/DeviceTable.tsx
+++ b/packages/ui/origin-ui-core/src/components/devices/Table/DeviceTable.tsx
@@ -102,7 +102,7 @@ export function DeviceTable(props: IOwnProps) {
 
     useEffect(() => {
         loadPage(1);
-    }, [user]);
+    }, [user, props.devices]);
 
     function viewDevice(rowIndex: number) {
         const device = paginatedData[rowIndex];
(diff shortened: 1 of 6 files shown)


josipbagaric:
Hmm, does this mean that meter reads were not fetched at all previously?

alexworker23:
> Hmm, does this mean that meter reads were not fetched at all previously?

Yes, they were fetched only when user opened DeviceDetailView. 
But this flow existed for a couple of weeks, after my Devices refactor.

josipbagaric:
> > Hmm, does this mean that meter reads were not fetched at all previously?
> 
> Yes, they were fetched only when user opened DeviceDetailView.
> But this flow existed for a couple of weeks, after my Devices refactor.

So now they are fetched as soon as you open the device tables?

alexworker23:
Now we have it like this:
1. Open Device table
2. Device data without meterReads is fetched and stored
3. User can see table with devices, but meterReads 0 
4. Fetching of devices + meterReads is starting
5. Device data with meterReads is received and stored - updating the view of the table.

josipbagaric:
> Now we have it like this:
> 
> 1. Open Device table
> 2. Device data without meterReads is fetched and stored
> 3. User can see table with devices, but meterReads 0
> 4. Fetching of devices + meterReads is starting
> 5. Device data with meterReads is received and stored - updating the view of the table.

Looks good, but we shouldn't show meter reads as **0**, but as **Loading**.
The user might be confused thinking that his meter reading is actually 0.

josipbagaric (approved):

PR #2215, opened 2021-03-23
Fix: Release creds

josipbagaric:
Potential fix for  https://github.com/energywebfoundation/origin/issues/2216

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,6 +73,13 @@ jobs:
         npx sleep 5
         yarn build:containers:release
       env:
-          GH_TOKEN: ${{ secrets.GITHUB_PERSONAL_ACCESS_TOKEN }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_PERSONAL_ACCESS_TOKEN }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          BUILD_ID: ${{ steps.generate_unique_id.outputs.unique_id }}
+          HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
+          HEROKU_STABLE_APP_API: ${{ secrets.HEROKU_STABLE_APP_API }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -72,3 +72,7 @@ jobs:
         yarn publish:release 
         npx sleep 5
         yarn build:containers:release
+      env:
+          GH_TOKEN: ${{ secrets.GITHUB_PERSONAL_ACCESS_TOKEN }}
```
GrzegorzMatuszakTSH: Maybe `GITHUB_TOKEN` var will do the job?

https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

josipbagaric: That's actually a good point. Possibly the issue is that GITHUB_TOKEN is not assigned to the action:

https://docs.github.com/en/actions/reference/authentication-in-a-workflow#example-passing-github_token-as-an-input

GrzegorzMatuszakTSH: I was thinking to set `${{ secrets.GITHUB_TOKEN }}` instead of `${{ secrets.GITHUB_PERSONAL_ACCESS_TOKEN }}` :)

```yaml
      env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          ```

josipbagaric: That shouldn't make a difference, `lerna publish` should pick up `secrets.GITHUB_TOKEN` automatically if it was available. But I think it isn't - check my link above.

Maybe I'm wrong 🤷‍♂️ not sure if there's a way we can test this PR before merging it

kosecki123:
Make sure you enable other ENV variables similarly to 

https://github.com/energywebfoundation/origin/blob/e27048e92806541624172d9e3001154c1d66a2a7/.github/workflows/deploy-master.yml#L77

.github/workflows/release.yml
josipbagaric: Actually, you might be right - I'll test it out.

josipbagaric:
Thanks @GrzegorzMatuszakTSH  @kosecki123. I've updated the commit and changed the target branch to `release`.

GrzegorzMatuszakTSH (approved):

PR #2223, opened 2021-03-23
chore: add I-REC UI Heroku image to push-heroku-stable

josipbagaric:


alexworker23 (commented):
I am not sure, if we should have it on stable or not

josipbagaric:
@kosecki123 any thoughts?

kosecki123 (approved):

PR #2297, opened 2021-04-08
feat/shared state

alexworker23:


packages/ui/origin-ui-v2/libs/ui/shared-state/src/configuration/localStorage.ts
```diff
@@ -0,0 +1,43 @@
...
+) => {
+  clearConfiguration();
+  await fetchAndSetConfiguration(options);
+};
+
+export const setPackageConfiguration: TFetchAndSetConfiguration = async (
+  options
+) => {
+  const existingConfig = getConfiguration();
+  if (!!existingConfig && !isEmpty(existingConfig)) {
```
arkadiusz-s: We dont need this extra check - isEmpty will suffice (undefined will return false

packages/ui/origin-ui-v2/libs/ui/shared-state/src/environment/fetch.ts
```diff
@@ -0,0 +1,17 @@
...
+  try {
+    const response = await fetch(customUrl ?? 'env-config.json');
+    const environment = await response.json();
+    return {
+      ...environment,
+      MARKET_UTC_OFFSET: parseInt(environment.MARKET_UTC_OFFSET, 10),
+    };
+  } catch (error) {
+    console.warn(
+      'Error while fetching env-config.json',
```
arkadiusz-s, line 15: We should probably prefix errors - maybe with (api)'  in this case  ( as in : "(api)   Error while fetching `env-config.json` file"

packages/ui/origin-ui-v2/libs/ui/shared-state/src/token/localStorage.ts
```diff
@@ -0,0 +1,18 @@
+import { AUTHENTICATION_TOKEN } from '../utils';
+import {
+  TGetAuthenticationToken,
+  TRemoveAutheticationToken,
+  TSetAuthenticationToken,
+} from './types';
+
+export const setAuthenticationToken: TSetAuthenticationToken = (token) => {
```
arkadiusz-s, line 8: We probably want additional param ( smth like `noReplace` which should probably be true by default ).

packages/ui/origin-ui-v2/libs/ui/shared-state/src/user/fetch.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const fetchUser: TFetchUser = async (options) => {
+  const { url, fetchFunc } = options;
+  try {
+    if (fetchFunc) {
+      return await fetchFunc();
+    }
+
+    const response = await fetch(url);
+    const user = await response.json();
```
arkadiusz-s, line 11: We might want to explicitly type payload

arkadiusz-s (changes requested):

packages/ui/origin-ui-v2/libs/ui/shared-state/src/token/localStorage.ts
arkadiusz-s: resolution -> not necessary.

arkadiusz-s (approved):

PR #2321, opened 2021-04-15
Move utils to UI utils package

arkadiusz-s:


packages/ui/origin-ui-v2/libs/ui/utils/src/exchange/fetchTotalVolume.ts
```diff
@@ -0,0 +1,34 @@
+import { DemandClient } from '@energyweb/exchange-irec-client';
+import { ICalculateVolumeData } from '../types';
+import { isEmpty } from 'lodash';
+import { EnergyFormatter } from '../formatters';
+import { toBN } from '../index';
+
+export async function fetchTotalVolume(
```
alexworker23, line 7: I think this will be a part of `exchange-data` package

packages/ui/origin-ui-v2/libs/ui/utils/src/types/index.ts
```diff
@@ -0,0 +1,332 @@
...
+  splits?: Split[];
+};
+
+export type BundleItemDTO = {
+  assetId: string;
+  volume: string;
+};
+
+export type CreateBundleDTO = {
+  price: number;
```
alexworker23, line 230: These types will be removed, and the types from appropriate `backend-api-client` package will be used

packages/ui/origin-ui-v2/libs/ui/validation/src/yup/setYupLanguage.ts
```diff
@@ -0,0 +1,61 @@
+import { setLocale } from 'yup';
+import printValue from 'yup/lib/util/printValue';
+import { TFunction } from 'i18next';
+
+export const setYupLanguage = (t: TFunction) => {
+  const translate = (path: string) => {
+    return (values: any) => t(`validations.yup.${path}`, values);
+  };
+  setLocale({
```
alexworker23, line 9: I am not sure if we will be using yup-specific localization, as long as we are using general localization

alexworker23 (approved):

PR #2337, opened 2021-04-19
chore(issuer-api): log errors when resyncing certificate fails

josipbagaric:
Log the error in the Logger when a certificate resync fails.

### DIFF
--- a/packages/traceability/issuer-api/src/pods/certificate/handlers/sync-certificate.handler.ts
+++ b/packages/traceability/issuer-api/src/pods/certificate/handlers/sync-certificate.handler.ts
@@ -27,7 +27,9 @@ export class SyncCertificateHandler implements IEventHandler<SyncCertificateEven
         try {
             await certificate.sync();
         } catch (e) {
-            this.logger.error(`Failed to resync certificate: ${JSON.stringify(e.message)}`);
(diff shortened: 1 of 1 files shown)


packages/traceability/issuer-api/src/pods/certificate/handlers/sync-certificate.handler.ts
```diff
@@ -27,6 +27,8 @@ export class SyncCertificateHandler implements IEventHandler<SyncCertificateEven
         try {
             await certificate.sync();
         } catch (e) {
+            this.logger.error(`Failed to resync certificate: ${JSON.stringify(e.message)}`);
```
soanvig: Don't we need information about tokenId, so we can debug which particullary certificate failed?
We have no guarantee, that such information will be in `e.message`, have we?

We can relate this with the `.log` in the beginning of this function, but that may be tedious to do.

josipbagaric: The error will probably include the tokenId + it's logged in the beginning of the function, but you're right - would be better to have it in the same log itself. Done.

soanvig (approved):

PR #2338, opened 2021-04-20
chore(docker): docker compose files to start 4 isolated variants of the Origin

artursudnik:
See docs/docker.md for details.

There is still a small risk of pulling energyweb/origin-backend-app:latest, energyweb/origin-backend-app:canary from Docker Hub instead of building them locally. This happens if these images are untagged by accident before building migrations images separately. If unnoticed, it leads to confusion and time wasted for debugging.


docs/docker.md
```diff
@@ -0,0 +1,72 @@
+# Running all the components with docker compose
+
+There are 4 docker compose files provided to run origin stack:
+
+-   based on the latest **stable** versions of NPM packages:
+    -   Vanilla Origin version with a **custom Device Registry** (`origin.release.yml`)
+    -   **I-REC compliant** Origin version (`origin.release.irec.yml`)
+-   based on the latest **canary** versions of NPM packages:
```
josipbagaric: I would just mention the Canary releases, the release version should never be run locally, but only on Heroku/AWS.

docs/docker.md
```diff
@@ -0,0 +1,72 @@
...
+rush update
+rush build #in case of errors it is necessary to execute rush rebuild
+```
+
+-   Build docker images
+
+    -   For the latest stable release:
+
+        ```shell
+        yarn build:containers:release
```
josipbagaric: This can also be removed from the README, as it will probably only be used by the CI/CD.

docs/docker.md
```diff
@@ -0,0 +1,72 @@
...
+        ```shell
+        yarn build:containers:canary
+        ```
+
+## Start chosen stack
+
+To start chosen stack, for example **I-REC compliant** Origin, execute the following:
+
+```shell
+docker-compose -f origin.release.irec.yml up -d
```
josipbagaric: Please use `origin.local.yml ` as the example YAML file in this file, as that one will probably be the only one that new devs will use. `origin.release.irec.yml` will only ever be used by the CI/CD.

josipbagaric (commented):
Looks good to me! Minor remarks regarding the README

docs/docker.md
artursudnik: do you mean building locally stable or canary images? It may be confusing if someone tries to run this setup without images built locally. Outdated versions will be fetched from the docker hub. If we do not plan to maintain them I would remove them from the hub.

josipbagaric: I meant just removing the instructions for stable from the README. Canary should be there, and that's the one that devs would be using for local development.

Devs shouldn't really run the stable releases in their environments, the CI/CD system should once deployed to a server, so I'm not sure if it's useful having it in the README.

josipbagaric (approved):

PR #2339, opened 2021-04-20
fix(issuer-api): Use getRepository when self-updating certificate

josipbagaric:
`Certificate.update` sometimes errors with the incorrect connection to the DB.
Instead use `getRepository(Certificate)` and update through the repository, to stay consistent with the rest of the codebase.


soanvig (approved):
I like better error handling

PR #2340, opened 2021-04-21
fix(issuer-api): Certificate events syncing

josipbagaric:
Fixes:
1. `TransferBatch` and `ClaimBatch` was listening to the wrong on-chain event
2. Remove double syncing certificates - only sync certificates when a blockchain event is detected
3. Sync certificates from an event instead of an internal `sync()` function
4. Wait for transactions to be mined before sending a confirmation

### DIFF
--- a/packages/traceability/issuer-api/src/pods/certificate/handlers/transfer-certificate.handler.ts
+++ b/packages/traceability/issuer-api/src/pods/certificate/handlers/transfer-certificate.handler.ts
@@ -64,16 +64,22 @@ export class TransferCertificateHandler implements ICommandHandler<TransferCerti
         }
 
         try {
-            await onChainCert.transfer(
+            const transferTx = await onChainCert.transfer(
                 to,
                 BigNumber.from(amount ?? onChainCert.owners[from]),
                 from
             );
+
+            const receipt = await transferTx.wait(1);
+
+            if (receipt.status === 0) {
+                throw new Error(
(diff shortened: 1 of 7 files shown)


packages/traceability/issuer-api/src/pods/certificate/handlers/transfer-certificate.handler.ts
```diff
@@ -64,17 +64,23 @@ export class TransferCertificateHandler implements ICommandHandler<TransferCerti
...
 
         try {
-            await onChainCert.transfer(
+            const transferTx = await onChainCert.transfer(
                 to,
                 BigNumber.from(amount ?? onChainCert.owners[from]),
                 from
             );
+
+            const receipt = await transferTx.wait(1);
```
soanvig: What's that wait(1)? It looks creepy

josipbagaric: Wait for 1 transaction confirmation for the mined block.

soanvig: For a sec I thought it's like 1ms or something :D Thanks god

soanvig:
I reviewed the changes, but those are too domain specific to approve them (for me, for now), so I leave it to @ioncreature

ioncreature (approved):

PR #2341, opened 2021-04-23
feat(issuer): Smart Contract Improvements

josipbagaric:
- [x] Use OpenZeppelin ERC-1155 instead of a custom implementation
- [x] Create a different smart contract for private certificate issuance `PrivateIssuer.sol`
- [x] Remove `encodeData` and `decodeData` from smart contracts, since we now have an off-chain implementation
- [x] Add `privateIssuer` contract as an optional contract to `BlockchainProperties`
- [x] ~1500 lines of code removed 🥳


packages/traceability/issuer-api/src/pods/blockchain/blockchain-properties.dto.ts
```diff
@@ -1,29 +1,21 @@
 import { ApiProperty } from '@nestjs/swagger';
-import { IsInt, IsNotEmpty, IsString } from 'class-validator';
 
 export class BlockchainPropertiesDTO {
     @ApiProperty({ type: Number })
-    @IsNotEmpty()
```
soanvig, line 6: Why validation is removed?

packages/traceability/issuer-api/src/pods/blockchain/blockchain-properties.entity.ts
```diff
@@ -26,10 +26,15 @@ export class BlockchainProperties extends ExtendedBaseEntity {
...
-    rpcNodeFallback: string;
+    rpcNodeFallback?: string;
+
+    @Column({ nullable: true })
+    privateIssuer?: string;
 
     wrap(signerOrPrivateKey?: Signer | string): IBlockchainProperties {
-        const web3 = getProviderWithFallback(this.rpcNode, this.rpcNodeFallback);
+        const web3 = getProviderWithFallback(
+            ...[this.rpcNode, this.rpcNodeFallback].filter((url) => !!url)
```
soanvig, line 36: Just a super NIT, but for that kind of functions it's nice to have a helper. lodash has `compact`.

For this case though maybe you consider `.filter(Boolean)` since it better explains intention? It's just my opinion though.

packages/traceability/issuer-api/test/issuer-api.ts
```diff
@@ -117,13 +121,16 @@ export class StubValidateDeviceOwnershipQueryHandler
...
+
+    await issuer.setPrivateIssuer(privateIssuer.address);
 
     const moduleFixture = await Test.createTestingModule({
         imports: [
             TypeOrmModule.forRoot({
                 type: 'postgres',
                 host: process.env.DB_HOST ?? 'localhost',
-                port: Number(process.env.DB_PORT) ?? 5432,
+                port: Number(process.env.DB_PORT ?? 5432),
```
soanvig, line 133: classic :D

packages/trade/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
```diff
@@ -90,7 +90,7 @@ export class DepositWatcherService implements OnModuleInit {
 
         const { transactionHash } = event;
 
-        const { _from: from, _to: to, _value: value, _id: id } = log;
+        const [, from, to, id, value] = log;
```
soanvig, line 93: Please use `_` for unused variable, this `,` is really cryptic

soanvig (commented):
Not really useful comments, but I don't understand what's happening here :/

packages/traceability/issuer-api/src/pods/blockchain/blockchain-properties.dto.ts
josipbagaric: Because it's never used as a "create" DTO, only as a "return" DTO. So the validation is never really triggered, thus useless.

packages/trade/exchange-io-erc1888/src/deposit-watcher/deposit-watcher.service.ts
josipbagaric: Deepscan will probably complain about the unused variable.

packages/traceability/issuer/contracts/Registry.sol
```diff
@@ -1,39 +1,41 @@
...
-import "./ERC1155/ERC1155Mintable.sol";
+import "@openzeppelin/contracts/token/ERC1155/ERC1155.sol";
 import "./ERC1888/IERC1888.sol";
 
-contract Registry is ERC1155Mintable, ERC1888 {
+contract Registry is ERC1155, ERC1888 {
 
 	mapping(uint256 => Certificate) public certificateStorage;
-	mapping(uint256 => mapping(address => uint256)) public claimedBalances;
+	mapping(uint256 => mapping(address => uint256)) public _claimedBalances;
```
kosecki123: Should this be without `_` since it's public ?

kosecki123 (approved):

PR #2342, opened 2021-04-23
React query api client generation

artursudnik:
This is a prototype of the code generation during a build. 

To be discussed:
- destination folder
- how to automatically import/export all modules generated
- how to avoid redefining mutator, providers, anything that is going to be shared between autogenerated clients


artursudnik:
Decisions made:
- destination folder: /packages/api-clients/react-query
- generating index.ts with a script after client .ts files are generated
- shared mutator for all clients

Still to be solved:
- setting backend baseUrl

artursudnik:
This is a prototype for exchange. If it works I will reproduce this pattern for all other clients.
Every client is going to have its own mutator instance. To be solved later if it is an issue.

packages/api-clients/react-query/exchange/src/mutator/customInstance.ts
```diff
@@ -0,0 +1,18 @@
+import Axios, { AxiosRequestConfig } from 'axios';
+
+export const AXIOS_INSTANCE = Axios.create({ baseURL: '' });
```
alexworker23: I think here as a temporary solution we should add `process.env.BACKEND_URL+process.env.BACKEND_PORT` or new single line `BACKEND_URL` env

artursudnik: I agree. I can imagine a situation when someone needs to expose backend API under some path. It is not possible to specify it now with these two variables. But, I'd introduce this change separately.

PR #2358, opened 2021-04-26
fix(origin-ui-core): check irec enabled feature in organizationMenu

soanvig:
Organization menu was unable to render if irec was disabled

### DIFF
--- a/packages/ui/origin-ui-core/src/components/organization/Organization.tsx
+++ b/packages/ui/origin-ui-core/src/components/organization/Organization.tsx
@@ -7,7 +7,8 @@ import { useLinks } from '../../hooks';
 export const Organization = () => {
     const { organizationPageUrl } = useLinks();
     const organizationMenuList = useOrganizationMenu();
-    const firstNotHiddenRoute = organizationMenuList.filter((i) => i.show)[0]?.key;
+    const displayableMenuList = organizationMenuList.filter((i) => i.show);
+    const firstNotHiddenRoute = displayableMenuList[0]?.key;
 
     return (
         <div className="PageWrapper">
@@ -15,7 +16,7 @@ export const Organization = () => {
                 path={`${organizationPageUrl}/:key/:id?`}
                 render={(props) => {
                     const key = props.match.params.key;
-                    const matches = organizationMenuList.filter((item) => {
+                    const matches = displayableMenuList.filter((item) => {
                         return item.key === key;
                     });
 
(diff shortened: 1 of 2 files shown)


packages/ui/origin-ui-core/src/components/organization/Organization.tsx
```diff
@@ -7,15 +7,16 @@ import { useLinks } from '../../hooks';
...
+    const firstNotHiddenRoute = displayableMenuList[0]?.key;
 
     return (
         <div className="PageWrapper">
             <Route
                 path={`${organizationPageUrl}/:key/:id?`}
                 render={(props) => {
                     const key = props.match.params.key;
-                    const matches = organizationMenuList.filter((item) => {
+                    const matches = displayableMenuList.filter((item) => {
```
soanvig, line 19: Here the match should work only with displayable items, not with all available (IMO)

packages/ui/origin-ui-core/src/components/organization/organizationMenu.ts
```diff
@@ -78,8 +79,9 @@ export const useOrganizationMenu = (): IOrganizationMenuItem[] => {
             key: 'register-irec',
             label: 'Register I-REC',
             component: IRECRegisterForm,
-            feature: OriginFeature.IRec,
             show:
+                config && // sometimes context returns null as config
```
soanvig: I hate that kind of comments, but this is the problem, that is hard to verify, due to problems @alexworker23  mentioned with rerendering.

If you have another idea, please share

packages/ui/origin-ui-core/src/components/organization/organizationMenu.ts
```diff
@@ -9,19 +9,20 @@ import { OrganizationForm } from './OrganizationForm';
...
 import { OrganizationView } from './OrganizationView';
 import { fromUsersSelectors } from '../../features';
+import { useOriginConfiguration } from '../../utils/configuration';
 
 interface IOrganizationMenuItem {
     key: string;
     label: string;
     component: React.ReactType;
     show: boolean;
-    feature?: OriginFeature;
```
soanvig, line 18: It wasn't used and IMO at this point it only duplicates logic `show` is doing in slightly more declarative way. However, because of `?` handling this would be more complex, so it's better to remove it for now

alexworker23 (approved):

packages/ui/origin-ui-core/src/components/organization/organizationMenu.ts
```diff
@@ -78,8 +79,9 @@ export const useOrganizationMenu = (): IOrganizationMenuItem[] => {
             key: 'register-irec',
             label: 'Register I-REC',
             component: IRECRegisterForm,
-            feature: OriginFeature.IRec,
             show:
+                config &&
```
soanvig, line 83: Sometimes context returned null. This may be due to some problems with rerendering and such

PR #2037, opened 2021-02-16
chore(exchange): bulk claim

josipbagaric:
In order to allow claiming from the exchange from the UI, we need to enable bulk claiming certificates directly from the exchange.

PR #2397, opened 2021-05-11
feat(exchange): Bulk claim from the exchange

josipbagaric:
Adds an option to the exchange to bulk claim certificates directly from the exchange.

_Note: For simplicity, it doesn't use bulk claiming function on the smart contracts - this would require us to change the exchange Transfer entity to accommodate bulk transfers. This could be a future improvement._

Continuation of the PR https://github.com/energywebfoundation/origin/pull/2037.


kosecki123 (approved):

soanvig:
I wonder if we need single claim  if we have bulk claim. This is effectively duplicating logic, ain't?

Of course for API purposes we may have separate endpoint, but underhood it can use "bulk" claim (which would be the default).  Single claim could be like a bulk claim, but with one entry only.

josipbagaric:
> I wonder if we need single claim if we have bulk claim. This is effectively duplicating logic, ain't?
> 
> Of course for API purposes we may have separate endpoint, but underhood it can use "bulk" claim (which would be the default). Single claim could be like a bulk claim, but with one entry only.

They function a bit differently, mainly because when you do a single claim you specify an amount you want to claim, but when you do bulk claims then you claim all available amounts for each of the certificates.

Good thoughts though. I've abstracted the logic a bit now.

soanvig (approved):

PR #2399, opened 2021-05-13
chore(issuer): MythX audit results

josipbagaric:
Adding minor changes to smart contracts after a MythX audit result.


kosecki123 (approved):
Is that only issues MythX found?

josipbagaric:
> Is that only issues MythX found?

Yup

soanvig (approved):

soanvig:
Is it possible to add this to ci so it verifies security on PR?

josipbagaric:
> Is it possible to add this to ci so it verifies security on PR?

Usually yes, but their CLI tool doesn't work with our codebase out-of-the-box. I had to fork their tool and change the source code to make it work. Will open an issue on their repo to fix it and then we can think about adding it to CI.

josipbagaric:
Shouldn't be a huge issue though as the smart contracts will rarely, if ever, change.

ioncreature (approved):

PR #2422, opened 2021-05-18
Chore/file upload

alexworker23:
Made FileUpload component work and add it to Organization Registration.
image
Added a download component to ui-core + added docs download list for Organization View
image


josipbagaric (approved):
LGTM, just not sure why the whitespace is 2 spaces instead of 4 spaces like elsewhere in the repo. Thought we had a linting rule for that?

alexworker23:
> LGTM, just not sure why the whitespace is 2 spaces instead of 4 spaces like elsewhere in the repo. Thought we had a linting rule for that?

I guess it happens, because NX has its own linting rules inside, which is not configured yet to match with the rest of repo.

PR #2454, opened 2021-05-27
fix(db-schema): Configuration.id definition

artursudnik:
current Configuration.id field definition causes the following changes added on every typeorm generated migration:
```
    public async up(queryRunner: QueryRunner): Promise<void> {   
        await queryRunner.query(`ALTER TABLE "configuration" ALTER COLUMN "id" SET DEFAULT 1`);
    }

    public async down(queryRunner: QueryRunner): Promise<void> {
        await queryRunner.query(`ALTER TABLE "configuration" ALTER COLUMN "id" SET DEFAULT '1'`);
    }
```


josipbagaric (approved):
👌

josipbagaric:
Does this stop the migration creation now?

artursudnik:
After this fix only intended changes are included in a created migration.

PR #2475, opened 2021-05-30
feat(origin-device-registry-irec-local-api): renamed irec device and fuels types

ioncreature:


packages/devices/origin-device-registry-irec-local-api/src/device/device.controller.ts
```diff
@@ -88,7 +88,7 @@ export class DeviceController {
         description: 'Returns all IREC fuels types'
     })
     getFuelTypes(): CodeNameDTO[] {
```
alexworker23: These controller functions (+ descriptions) should be renamed as well, because their naming is later transferred to auto-generated clients.

packages/devices/origin-device-registry-irec-local-api/src/device/device.controller.ts
```diff
@@ -76,7 +76,7 @@ export class DeviceController {
         description: 'Returns all IREC fuels'
     })
     getFuels(): CodeNameDTO[] {
-        const deviceTypes = this.deviceService.getDeviceTypes();
+        const deviceTypes = this.deviceService.getFuelTypes();
```
alexworker23: I think it is a mismatch. `this.deviceService.getDeviceTypes()` should be used here

ioncreature: done

packages/traceability/issuer-irec-api-wrapper/src/Device.ts
```diff
@@ -29,11 +29,11 @@ export class DeviceCreateParams extends FileIds {
 
     @Expose({ name: 'device_type_code', toPlainOnly: true })
     @IsString()
-    deviceType: string;
+    fuelType: string;
```
alexworker23, line 32: Mismatch between `@Expose({ name: 'device_type_code', toPlainOnly: true })`  and `fuelType: string`.
Same goes for line 36, line 99, line 104 in this file.

packages/traceability/issuer-irec-api-wrapper/src/Issue.ts
```diff
@@ -23,7 +23,7 @@ export class Issue extends FileIds {
     @Expose({ name: 'fuel_code', toPlainOnly: true })
     @IsString()
     @IsNotEmpty()
-    fuel: string;
+    deviceType: string;
```
alexworker23, line 26: Mismatch between `@Expose({ name: 'fuel_code', toPlainOnly: true })` and `deviceType: string`.

packages/traceability/issuer-irec-api-wrapper/src/Device.ts
ioncreature: it is not a mismatch. IREC API expects values with "TC" prefixes here. And we want "TC" to be a fuel type. So I changed it since we want to use different from IREC terminology

ioncreature: see here
https://evident-registry.atlassian.net/wiki/spaces/EV/pages/219447320/POST%2Bapi%2Birec%2Bv1%2Bdevice-management%2Bcreate

packages/traceability/issuer-irec-api-wrapper/src/Issue.ts
ioncreature: same as above

PR #2479, opened 2021-05-31
chore(issuer-api): Unify certificate and certification request IDs

josipbagaric:
- [x] Use `certificate.tokenId` as `certificate.id` and remove tokenId
- [x] Use `certificationRequest.requestId` as `certificationRequest.id` and remove requestId
- [x] Make sure migrations work (tested on a cloned database from canary - works fine)
- [ ] ~Handle cases where there is a possibility of nonce conflicts~ (Will create a new PR for this)


packages/traceability/issuer-api/migrations/1622458600677-RemoveTokenIds.ts
```diff
@@ -0,0 +1,119 @@
...
+        await queryRunner.query(`DROP SEQUENCE "issuer_certification_request_id_seq"`);
+
+        await Promise.all(
+            oldCertificationRequests.map(async (oldCertReq: any) => {
+                return queryRunner.query(
+                    `UPDATE "issuer_certification_request" SET id = ${oldCertReq.requestId} WHERE id = '${oldCertReq.id}'`
+                );
+            })
+        );
+    }
```
ioncreature: Dangerous approach, we can lose all `tokenId`'s and down migration won't help.
Let's better first copy `tokenId`'s to 'id' and then drop `tokenId` column

packages/traceability/issuer-api/migrations/1622458600677-RemoveTokenIds.ts
```diff
@@ -0,0 +1,119 @@
...
+            `ALTER TABLE "issuer_certification_request" ADD CONSTRAINT "UQ_551869cc9ee5caeccd53c966cdd" UNIQUE ("requestId")`
+        );
+
+        await Promise.all(
+            oldCertificationRequests.map(async (oldCertReq: any) => {
+                return queryRunner.query(
+                    `UPDATE "issuer_certification_request" SET "requestId" = ${oldCertReq.requestId} WHERE id = '${oldCertReq.id}'`
+                );
+            })
+        );
```
ioncreature: same as above

packages/traceability/issuer-irec-api/src/pods/certification-request/handlers/create-irec-certification-request.handler.ts
```diff
@@ -5,17 +5,17 @@ import { createReadStream } from 'fs';
...
+
 import { CreateIrecCertificationRequestCommand } from '../commands';
 import { IrecCertificateService } from '../irec-certificate.service';
 import { FullCertificationRequestDTO } from '../full-certification-request.dto';
 import { IrecCertificationRequest } from '../irec-certification-request.entity';
 
 @CommandHandler(CreateIrecCertificationRequestCommand)
-export class CreateCertificationRequestHandler
-    extends OriginalHandler
+export class CreateIrecCertificationRequestHandler
```
ioncreature, line 17: Why did you change the name?

ioncreature: It probably wont work with `origin-irec-backend`

ioncreature (changes requested):

packages/traceability/issuer-api/migrations/1622458600677-RemoveTokenIds.ts
soanvig: We should consider not dropping or rewriting columns at all, but just rename columns around and in future remove old ones. In this case this would rename `id` to `id_bak`, then rename `requestId` to `id`

packages/traceability/issuer-api/migrations/1622458600677-RemoveTokenIds.ts
soanvig: and the same as above in my case as well

packages/traceability/issuer-api/migrations/1622459627494-RmCertificationRequestStatus.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class RmCertificationRequestStatus1622459627494 implements MigrationInterface {
+    name = 'RmCertificationRequestStatus1622459627494';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "issuer_certification_request" DROP COLUMN "status"`);
```
soanvig, line 7: This is some additional change?

packages/traceability/issuer-api/src/pods/certification-request/certification-request.controller.ts
```diff
@@ -122,7 +122,7 @@ export class CertificationRequestController {
             new ValidateCertificationRequestCommand(dto)
         );
 
-        if (!validationCheck.success) {
+        if (!validationCheck?.success) {
             return validationCheck;
```
soanvig: This code doesn't make sense now - if validationCheck can be undefined, and THEN we enter the if, the undefined will be returned

soanvig (changes requested):

packages/traceability/issuer-api/migrations/1622458600677-RemoveTokenIds.ts
josipbagaric: Good point @ioncreature. It's an edge case, since tokenIds are stored in memory, but a valid one.

@soanvig I would consider this only in cases where there's no clear `down()` migrations. Here it's unnecessary, as the `down()` migration will work seamlessly with existing data.

packages/traceability/issuer-irec-api/src/pods/certification-request/handlers/create-irec-certification-request.handler.ts
josipbagaric: Changed the name because it works differently now - it calls the original handler and supplements it with I-REC features. It doesn't override the original handler.

Why won't it work? I'll do more thorough testing on this.

packages/traceability/issuer-api/src/pods/certification-request/certification-request.controller.ts
josipbagaric: Good point, didn't pay too much attention to this just wanted it to compile. Will rework this part.

packages/traceability/issuer-api/migrations/1622459627494-RmCertificationRequestStatus.ts
josipbagaric: Yup. Certification requests no longer need the `status` field because they are stored in the db only once they are deployed to the blockchain (they are no longer queued - this queue will be replaced with an application wide-queue).

packages/traceability/issuer-api/src/pods/certification-request/certification-request.controller.ts
josipbagaric: Undid the change for now, as it's not needed it seems. Probably pushed by mistake.

packages/traceability/issuer-irec-api/src/pods/certification-request/handlers/create-irec-certification-request.handler.ts
ioncreature: It could not work because this package is like a proxy for `issuer-api`. And `origin-backend-irec-app` works only with this package, and not with `issuer-api` anymore. 
So since this package is a proxy, it has to have the same signatures with `issuer-api` everywhere.

packages/traceability/issuer-api/migrations/1622458600677-RemoveTokenIds.ts
```diff
@@ -0,0 +1,120 @@
...
+        await queryRunner.query(`ALTER TABLE "issuer_certificate" ALTER COLUMN "id" DROP DEFAULT`);
+        await queryRunner.query(`DROP SEQUENCE "issuer_certificate_id_seq"`);
+
+        await Promise.all(
+            oldCertificates.map(async (oldCertificate: any) => {
+                return queryRunner.query(
+                    `UPDATE "issuer_certificate" SET id = ${oldCertificate.tokenId} WHERE id = '${oldCertificate.id}'`
+                );
+            })
+        );
```
ioncreature: 1. Let's do `DROP SEQUENCE IF EXISTS "issuer_certificate_id_seq"`. This is for the case when we run the same migration more than once (eg it failed on the first run)
2. `await Promise.all()` looks like we are ddosing our own DB. Let's rewrite it with `for` loop or with longer SQL query or with batch SQL query

ioncreature (changes requested):

packages/traceability/issuer-irec-api/src/pods/certification-request/handlers/create-irec-certification-request.handler.ts
josipbagaric: How do I test this flow? I see no automatic tests in the package.

ioncreature: Manually only, I'll add tests later

soanvig (approved):
Don't trust my approval too much though

packages/traceability/issuer-irec-api/src/pods/certification-request/handlers/create-irec-certification-request.handler.ts
ioncreature: added tests

ioncreature (approved):

PR #2482, opened 2021-06-01
feat(origin-organization-irec-api): beneficiary API

ioncreature:
- [x] IREC module
- [x] Beneficiary module with API controller
- [x] Beneficiary API tests


packages/organizations/origin-organization-irec-api/src/beneficiary/handlers/create-beneficiary.handler.ts
```diff
@@ -0,0 +1,40 @@
...
+
+        const irecBeneficiary = await this.irecService.createBeneficiary(
+            platformAdmin.id,
+            organization
+        );
+
+        const beneficiary = this.repository.create({
+            irecBeneficiaryId: irecBeneficiary.id,
+            organizationId: organization.id,
+            ownerOrganizationId: platformAdmin.organization.id
```
kosecki123: Hey @ioncreature what's the idea behind that ownerOrganizationId field? Seems like that's always set to a platform owner organization.

packages/organizations/origin-organization-irec-api/src/beneficiary/dto/beneficiary.dto.ts
```diff
@@ -0,0 +1,21 @@
...
+    id: number;
+
+    @ApiProperty({ type: Number })
+    irecBeneficiaryId: number;
+    @ApiProperty({ type: PublicOrganizationInfoDTO })
+    @Type(() => PublicOrganizationInfoDTO)
+    organization: PublicOrganizationInfoDTO;
+
+    @ApiProperty({ type: Number })
+    ownerOrganizationId: number;
```
kosecki123: If this is mean to be the organization that owns the beneficiary record, I suggest renaming this field to `ownerId` as it's used in other parts of the system.

packages/organizations/origin-organization-irec-api/src/beneficiary/handlers/add-organization-beneficiary.handler.ts
```diff
@@ -0,0 +1,47 @@
...
+        private readonly repository: Repository<Beneficiary>,
+        private readonly userService: UserService,
+        private readonly commandBus: CommandBus
+    ) {}
+
+    public async execute({
+        ownerOrganizationId,
+        irecBeneficiaryId
+    }: AddOrganizationBeneficiaryCommand): Promise<BeneficiaryDTO> {
+        const platformAdmin = await this.userService.getPlatformAdmin();
```
kosecki123: If `ownerOrganizationId` is meant to be the id of the organization that the user belongs to then getPlatformAdmin is the wrong method as it returns the organization that owns the platform, not the users' organization.

packages/organizations/origin-organization-irec-api/src/beneficiary/handlers/create-beneficiary.handler.ts
ioncreature: We create a beneficiary for each new organization for the platform account.
But different organization will create their own subset of beneficiaries based on this field

packages/organizations/origin-organization-irec-api/src/beneficiary/handlers/add-organization-beneficiary.handler.ts
ioncreature: see above

packages/organizations/origin-organization-irec-api/src/beneficiary/dto/beneficiary.dto.ts
ioncreature: changed

packages/organizations/origin-organization-irec-api/src/beneficiary/handlers/create-beneficiary.handler.ts
kosecki123: `platformAdmin.organization.id` is always the same organization, like PTT or FOTON. Did you maybe meant to use `loggedInUser.organizationId` instead?

josipbagaric: I think `loggedInUser.organizationId` is stored in the line above as `organizationId`.

@ioncreature if `platformAdmin.organization.id` is needed, I assume there no need to store it in the database with every row, as it will always stay the same throughout the life-cycle of the application?

Can't it be fetched only when we interact with the I-REC wrapper?

kosecki123: @ioncreature I think we do a little bit of guessing here, could you please explain what each of these fields relates to?

ioncreature: @JosephBagaric Yep, agree. We can get rid of it. I will set `null` instead of `platformAdmin.organization.id`. And it will mean the same

josipbagaric: You can remove the entire column, no sense in keeping a column that's always `null`.

ioncreature: `irecBeneficiaryId` and `organizationId` identify IREC beneficiary and connected origin organization. So they are always together.

`ownerId` (was `ownerOrganizationId` before) is to show who is the owner.

It means that everybody can have read access to the platform beneficiaries list, but if someone wants to use beneficiary then he has to add this beneficiary to his own organization list. And then use.

ioncreature: https://github.com/energywebfoundation/origin/pull/2482#discussion_r648078093
Yes, we create all IREC beneficiaries from the platform account

ioncreature: So we have 2 different handlers - `create beneficiary` which creates beneficiary for org using platform account. 
And `add beneficiary` which adds it to the organization's private beneficiaries list

kosecki123: Given all that we discussed, I think the database table is correct so you store `irecBeneficiaryId`, `organizationId`, and `ownerId`. Since looks like IREC API beneficiary is bound to the IREC organization, you need to follow that principle as well here, so 

- each platform organization (ownerId) maintains it's own list of beneficiaries and the list is private to them
- during redemption (claiming) the organization who requests uses the `irecBeneficiaryId` that is created by them as `beneficiary_id` in https://evident-registry.atlassian.net/wiki/spaces/EV/pages/216531743/POST%2Bapi%2Birec%2Bv1%2Bredemption-management

ioncreature: @kosecki123 It is already it. Except the redemption - it is on different task

packages/organizations/origin-organization-irec-api/src/beneficiary/beneficiary.controller.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Get('/company')
+    @UseGuards(AuthGuard())
+    @ApiResponse({
+        status: HttpStatus.OK,
+        type: [BeneficiaryDTO],
+        description: 'Get company beneficiaries'
+    })
+    public async removeOrganizationBeneficiary(
```
kosecki123: Method name is wrong, seems like swaped with the real DELETE endpoint.

packages/organizations/origin-organization-irec-api/src/beneficiary/handlers/add-organization-beneficiary.handler.ts
```diff
@@ -0,0 +1,43 @@
...
+    ) {}
+
+    public async execute({
+        ownerId,
+        irecBeneficiaryId
+    }: AddOrganizationBeneficiaryCommand): Promise<BeneficiaryDTO> {
+        const beneficiary = await this.repository.findOne({
+            ownerId: null,
+            irecBeneficiaryId
+        });
```
kosecki123, line 27: In order to add beneficiaries to your organization list, it has to exist first in the DB (with ownerId=null), what's the process of filling that?

packages/organizations/origin-organization-irec-api/src/beneficiary/beneficiary.controller.ts
ioncreature: @kosecki123 what is your proposal?

ioncreature: fixed, mixed names `removeOrganizationBeneficiary` and `getCompanyBeneficiaries`
thanks

packages/organizations/origin-organization-irec-api/src/beneficiary/handlers/add-organization-beneficiary.handler.ts
ioncreature: 1. The user gets all the public beneficiaries (with ownerId=null)
2. The user adds beneficiaries to his organization's private beneficiary list. And set ownerId to his organization id
3. When the beneficiary is not needed anymore, it is deleted from the private beneficiaries list by `DELETE /irec/beneficiary/:id`

kosecki123: What's the process of getting a public list of beneficiaries?

ioncreature: `GET /irec/beneficiary` returns list of public beneficiaries
`GET /irec/beneficiary/company` returns list of private ones

kosecki123 (approved):

PR #2496, opened 2021-06-07
chore(issuer-api): Skip issuance TransferSingle event

josipbagaric:
When a certificate is [created](https://github.com/energywebfoundation/origin/blob/master/packages/traceability/issuer/contracts/Registry.sol#L16-L30) in the smart contract, two events are fired:
1. [IssuanceSingle](https://github.com/energywebfoundation/origin/blob/master/packages/traceability/issuer/contracts/Registry.sol#L29) event, designating that a certificate has been issued
2. [TransferSingle](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC1155/ERC1155.sol#L286) event, with the `from` address set to `0x00000000000000`, designating that a certificate has been minted.

We have to skip handling the 2nd event, so that we don't double handle issuance events.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -2,17 +2,17 @@ dependencies:
   '@babel/core': 7.14.3
   '@date-io/moment': 1.3.13_moment@2.29.1
   '@energyweb/energy-api-influxdb': 0.7.0_dc7550b32b1014ff62ea7cebd522ef42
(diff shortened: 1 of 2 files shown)


soanvig (approved):

soanvig:
To add more context: in some implementations there was race condition between those two events, and handler of event resulting from `TransferSingle` tried to access certificate that wasn't yet created by `IssuanceSingle`

PR #2500, opened 2021-06-11
feat(switchboard): DID organisation registration prototype

artursudnik:
To be implemented:
- [ ] saving blockchain-related data (to be decided, where)
- [ ] some of the fields can not be published on-chain and will need to be entered by a user on the frontend
- [ ] frontend organisation registration form using `organisation/did` endpoint
- [ ] design of the fronted flow
- [ ] tests


.env.example
```diff
@@ -22,7 +22,7 @@ DISABLED_UI_FEATURES=IRecConnect;DevicesImport;IRecUIApp;CertificatesImport
 
 # Secret keys
 DEPLOY_KEY=d9066ff9f753a1898709b568119055660a77d9aae4d7a4ad677b8fb3d2a571e5
-
+PLATFORM_OPERATOR_PRIVATE_KEY=e1e314bf0dac736ac81bb2c541ab4f6cf3a6de078045217dd10e1a07e673596f
```
josipbagaric, line 25: I think we currently use `DEPLOY_KEY` as the platform private key already.

PR #2518, opened 2021-06-18
feat: Add support for batch transactions on the API

josipbagaric:
Support batch transactions on the Issuer API:
- Batch Issuance
- Batch Transfer
- Batch Claiming

BREAKING CHANGE: Batch operations moved to the `/api/certificate-batch`

_Note_: Renamed "bulk" to "batch" to use a single terminology throughout the packages.


soanvig:
Could it be possible to make batch issue + transfer or issue + claim? Just wondering

josipbagaric:
Theoretically yes, but it's not a standard interface in ERC-1888, and it would add to the complexity of the smart contracts.

josipbagaric:
We could add an API endpoint to trigger both, but it would still be 2 transactions underneath and the request could take quite long.

Don't see a good enough reason yet to support such combos.

packages/traceability/issuer-api/src/pods/certificate/commands/batch-claim-certificates.dto.ts
```diff
@@ -0,0 +1,18 @@
+import { IntUnitsOfEnergy } from '@energyweb/origin-backend-utils';
+import { IClaimData } from '@energyweb/issuer';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { IsArray, IsObject, Validate } from 'class-validator';
+
+export class BatchClaimCertificatesDTO {
+    @ApiProperty({ type: [Number] })
+    @IsArray()
```
soanvig: You are not validating array content.

You can make it by using `@IsNumber({ each: true })` as far as I remember:

https://github.com/typestack/class-validator#validating-arrays

packages/traceability/issuer-api/src/pods/certificate/commands/batch-transfer-certificates.dto.ts
```diff
@@ -0,0 +1,17 @@
+import { IntUnitsOfEnergy } from '@energyweb/origin-backend-utils';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { IsArray, Validate, IsEthereumAddress } from 'class-validator';
+
+export class BatchTransferCertificatesDTO {
+    @ApiProperty({ type: [Number] })
+    @IsArray()
```
soanvig: The same

packages/traceability/issuer-api/src/pods/certificate/handlers/batch-issue-certificates.handler.ts
```diff
@@ -0,0 +1,36 @@
...
+import { BlockchainPropertiesService } from '../../blockchain/blockchain-properties.service';
+import { BatchIssueCertificatesCommand } from '../commands/batch-issue-certificates.command';
+
+@CommandHandler(BatchIssueCertificatesCommand)
+export class BatchIssueCertificatesHandler
+    implements ICommandHandler<BatchIssueCertificatesCommand>
+{
+    constructor(private readonly blockchainPropertiesService: BlockchainPropertiesService) {}
+
+    async execute({ certificatesInfo }: BatchIssueCertificatesCommand): Promise<ISuccessResponse> {
```
soanvig: This may be a problem. Singular implementation returns original certificate, which is usually used to get the certificate `id`. And this returns ISuccessResponse -> this way we completely lose track of the certificate

packages/traceability/issuer-api/src/pods/certificate/commands/batch-transfer-certificates.command.ts
```diff
@@ -0,0 +1,10 @@
+import { BigNumber } from 'ethers';
+
+export class BatchTransferCertificatesCommand {
+    constructor(
+        public readonly certificateIds: number[],
+        public readonly to: string,
+        public readonly forAddress: string,
+        public readonly values?: BigNumber[]
```
soanvig: This list is related to the list of certificateIds?

1. is length ensured to be the same length as certificateIds?
2. what if some value is missing (or is undefined in the middle of the list for example)?

To be honest I don't like that API. Wouldn't it better to accept tuples of certificateId and value?

packages/traceability/issuer-api/src/pods/certificate/commands/batch-transfer-certificates.command.ts
```diff
@@ -0,0 +1,10 @@
+import { BigNumber } from 'ethers';
+
+export class BatchTransferCertificatesCommand {
+    constructor(
+        public readonly certificateIds: number[],
+        public readonly to: string,
```
soanvig, line 6: So I have to transfer all certificates from one address to another? Isn't it possible to transfer different certificates to different addresses in one batch?

It will make fix complicated on implementation side, because certificates would have to additionally grouped into batches by this address.

packages/traceability/issuer-api/src/pods/certificate/handlers/batch-transfer-certificates.handler.ts
```diff
@@ -0,0 +1,68 @@
...
+        const certificatesToTransfer = await this.repository.findByIds(certificateIds, {
+            relations: ['blockchain']
+        });
+
+        if (
+            !certificatesToTransfer.every(
+                (cert) => cert.owners[forAddress] && BigNumber.from(cert.owners[forAddress]).gt(0)
+            )
+        ) {
+            return ResponseSuccess('You have requested claiming of a certificate you do not own');
```
soanvig: I guess you can easily return the list of certificates ids that you don't own to simplify debugging (sending 10 transfers, and one of them is wrong, requires me to debug 10 times more)

BTW this is transfer, not claim

soanvig (changes requested):

packages/traceability/issuer-api/src/pods/certificate/commands/batch-transfer-certificates.command.ts
josipbagaric: Yes it is.

1. It's ensured on the smart contract level.
2. The smart contract will reject it.

But you are right, tuples would be better. I just blindly followed the Solidity implementation which doesn't have tuples.

soanvig: I mean, object probably instead of tuple tuple :D

soanvig: Because I can make a transfer for one certificate to transfer all the energy (undefined value), and for another one only part, right

packages/traceability/issuer-api/src/pods/certificate/commands/batch-transfer-certificates.command.ts
josipbagaric: Unfortunately ERC-1888 doesn't support this. It only supports batch transferring from one address to another.
Same for batch claiming.

packages/traceability/issuer-api/test/certificate.e2e-spec.ts
```diff
@@ -117,8 +123,18 @@ describe('Certificate tests', () => {
         expect(certificate1.energy.publicVolume).to.equal(certificateTestData.energy);
     });
 
+    it.only('should batch issue certificates', async () => {
```
soanvig, line 126: nice :D

soanvig (changes requested):

packages/traceability/issuer-api/test/certificate.e2e-spec.ts
soanvig: image

packages/ui/origin-ui-core/src/components/Modal/ClaimModal.tsx
```diff
@@ -109,7 +108,13 @@ export function ClaimModal(props: IProps) {
         };
 
         const action = isBulkClaim
-            ? requestClaimCertificateBulk({ certificateIds, claimData })
+            ? requestClaimCertificateBulk({
+                  certificateAmounts: certificateIds.map((id) => ({
+                      id,
+                      amount: '0' // TO-DO: Replace 0 with a specific amount
```
soanvig: is that OK?

soanvig (approved):

packages/ui/origin-ui-core/src/components/Modal/ClaimModal.tsx
josipbagaric: ~It's OK for now since it will have to change with the UI refactor anyway.~

josipbagaric: Actually, I could change the `amount` to be of type `number | TotalOwnedAmount`.

josipbagaric: would probably need to write a custom validator for that

josipbagaric: Addressed this.

soanvig (approved):

PR #2539, opened 2021-06-20
feat(origin-device-registry-irec-local-api): default account for irec device creation is platform acc now

ioncreature:
- [x] Default account is platform account on device creation
- [x] IREC service refactoring
- [x] Device corn job moved to the device package


packages/devices/origin-device-registry-irec-local-api/src/device/device.module.ts
```diff
@@ -1,27 +1,37 @@
...
 import { DeviceController } from './device.controller';
 import { Device } from './device.entity';
 import { DeviceService } from './device.service';
 import { ValidateDeviceOwnershipCommandHandler } from './handlers/validate-device-ownership.handler';
-import { IrecDeviceService } from './irec-device.service';
+import { CheckDeviceStateTask } from './cron';
 
 @Module({
     imports: [
+        ScheduleModule.forRoot(),
```
soanvig: I guess this should be initialized in root module. As it is "forRoot" 😆 

No, but TBH I'm not sure if now the scheduling system will be available in neighbour module branches. I guess this should really by in root module

packages/organizations/origin-organization-irec-api/src/connection/connection.module.ts
```diff
@@ -3,23 +3,26 @@ import { CqrsModule } from '@nestjs/cqrs';
...
 import { RegistrationModule } from '../registration';
 import { IrecModule, IrecService } from '../irec';
 import { ConnectionController } from './connection.controller';
 import { Connection } from './connection.entity';
 import { ConnectionHandlers } from './handlers';
+import { RefreshAllTokensTask } from './cron';
 
 @Module({
     imports: [
+        ScheduleModule.forRoot(),
```
soanvig: Then if you apply this in root you don't have to import it here

packages/organizations/origin-organization-irec-api/src/irec/irec.service.ts
```diff
@@ -104,4 +115,203 @@ export class IrecService {
         const irecClient = await this.getIrecClient(user);
         return await irecClient.beneficiary.update(beneficiaryId, params);
     }
+
+    async createDevice(user: UserIdentifier, deviceData: DeviceCreateParams): Promise<IrecDevice> {
+        if (!this.isIrecIntegrationEnabled()) {
```
soanvig: Maybe this is beyond this scope, but IMHO this service is implemented in an ugly way.

If each method is replaced by the "isIrecIntegrationEnabled", then perhaps you should have separate service for that that would be **injected** instead of this primary one IF irec integration is disabled. 

it would make things a lot, a lot cleaner. To make things easier, IrecDisabledService could implement public methods of this IrecService, so they are always in sync.

soanvig (changes requested):

packages/devices/origin-device-registry-irec-local-api/src/device/device.module.ts
soanvig: root module of the package*

ioncreature: done

packages/organizations/origin-organization-irec-api/src/connection/connection.module.ts
ioncreature: done

packages/organizations/origin-organization-irec-api/src/irec/irec.service.ts
ioncreature: done, thanks for the good idea

packages/organizations/origin-organization-irec-api/src/irec/irec.module.ts
```diff
@@ -1,12 +1,29 @@
 import { Module } from '@nestjs/common';
+import { ConfigModule, ConfigService } from '@nestjs/config';
+import { CommandBus, CqrsModule } from '@nestjs/cqrs';
+
 import { IrecService } from './irec.service';
-import { ConfigModule } from '@nestjs/config';
-import { CqrsModule } from '@nestjs/cqrs';
+import { IrecMockService } from './irec.mock.service';
+
+export const IREC_SERVICE = Symbol('IREC Service');
```
soanvig, line 8: Nice! But was it neccessary? Have you tried to use `IrecService` class as a provider token? Just asking

packages/organizations/origin-organization-irec-api/src/irec/irec.service.ts
```diff
@@ -54,7 +98,7 @@ export class IrecService {
     }
 
     isIrecIntegrationEnabled(): boolean {
```
soanvig: is it used anywhere anymore?

packages/organizations/origin-organization-irec-api/src/irec/irec.module.ts
ioncreature: Yep, it is
It does not work without this. It loads real IrecService when I tried to use class reference instead of Symbol

soanvig: Interesting... I will check this in free time

soanvig (approved):

packages/organizations/origin-organization-irec-api/src/irec/irec.service.ts
ioncreature: removed

PR #2561, opened 2021-06-28
chore: Renovate - group solc updates [skip-ci]

josipbagaric:


soanvig (approved):
I don't know, this is questionable change. Risky as hell, and very insecure.

We should make a 2 hours meeting discussing this change.

:)

PR #2562, opened 2021-06-28
chore(issuer): Adjust to latest changes in ERC-1888 spec

josipbagaric:
Update to the latest ERC-1888 spec.

### DIFF
--- a/packages/traceability/issuer/contracts/Registry.sol
+++ b/packages/traceability/issuer/contracts/Registry.sol
@@ -7,6 +7,7 @@ import "./ERC1888/IERC1888.sol";
 
 /// @title Implementation of the Transferable Certificate standard ERC-1888.
 /// @dev Also complies to ERC-1155: https://eips.ethereum.org/EIPS/eip-1155.
+/// @dev ** Data set to 0 because there is no meaningful check yet to be done on the data
 contract Registry is ERC1155, ERC1888 {
 
 	// Storage for the Certificate structs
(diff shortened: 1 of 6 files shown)


soanvig (approved):

packages/traceability/issuer/contracts/Registry.sol
```diff
@@ -18,98 +19,93 @@ contract Registry is ERC1155, ERC1888 {
...
-	function issue(address _to, bytes calldata _validityData, uint256 _topic, uint256 _value, bytes calldata _issuanceData) external override returns (uint256) {
+	function issue(address _to, bytes calldata _validityData, uint256 _topic, uint256 _value, bytes calldata _data) external override returns (uint256 id) {
 		require(_to != address(0x0), "Registry::issue: to must be non-zero.");
 		
 		_validate(_msgSender(), _validityData);
 
-		uint256 id = ++_latestCertificateId;
-		ERC1155._mint(_to, id, _value, _issuanceData);
+		id = ++_latestCertificateId;
+		ERC1155._mint(_to, id, _value, new bytes(0)); // Check **
```
soanvig, line 32: So it's ERC-1888 but using ERC1155._mint method here? I guess it's not related, just asking

josipbagaric: ERC-1888 interface is an extension of the ERC-1155 interface. `Registry.sol` is thus an implementation of both ERC-1888 and ERC-1155.

https://github.com/energywebfoundation/origin/blob/master/packages/traceability/issuer/contracts/ERC1888/IERC1888.sol#L6

PR #2573, opened 2021-07-02
feat(irec-issuer-api): added irec certificate dto

ioncreature:
- [x] irec certificate dto
- [x] irec wrapper certificates schema update

### DIFF
--- a/packages/traceability/issuer-irec-api-wrapper/src/IRECAPIClient.ts
+++ b/packages/traceability/issuer-irec-api-wrapper/src/IRECAPIClient.ts
@@ -281,7 +281,7 @@ export class IRECAPIClient extends EventEmitter {
 
                 await validateOrReject(beneficiaryParams);
 
-                const url = `${beneficiaryManagementUrl}/${id}/edit`;
+                const url = `${beneficiaryManagementUrl}/${id}`;
 
                 const response = await this.axiosInstance.put<unknown>(
                     url,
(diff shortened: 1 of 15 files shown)


packages/traceability/issuer-irec-api-wrapper/src/IRECAPIClient.ts
```diff
@@ -482,13 +482,16 @@ export class IRECAPIClient extends EventEmitter {
     }
 
     public async redeem(redemption: Redemption): Promise<RedeemTransactionResult> {
+        const r =
+            redemption instanceof Redemption ? redemption : plainToClass(Redemption, redemption);
```
soanvig, line 486: I wonder - why it's required? It messed the code.

Theoretically class will be JSON.stringified by axios which will remove stuff like functions, and will use all strigifiable fields by default.

You can customize this behavior by using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description (`.toJSON`) - which I have mixed opinion about.

Also, if interface is not giant, it may be good idea to just manually  select all the fields you want to send.

ioncreature: Honestly, I do not understand you

If you ask why we use `plainToClass`, we need to apply data transforms for things like dates.

soanvig (approved):

PR #2598, opened 2021-07-06
fix: irec create certification requests issues

ioncreature:
several small issues on crons and irec certification requests

### DIFF
--- a/packages/devices/origin-device-registry-irec-local-api/src/device/cron/check-device-state.task.ts
+++ b/packages/devices/origin-device-registry-irec-local-api/src/device/cron/check-device-state.task.ts
@@ -20,7 +20,7 @@ export class CheckDeviceStateTask {
         const devices = await this.deviceService.findAll();
         for (const device of devices) {
             const irecDevice = await this.irecService.getDevice(device.ownerId, device.code);
-            if (irecDevice.status !== device.status) {
+            if (irecDevice && irecDevice.status !== device.status) {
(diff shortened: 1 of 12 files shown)


packages/devices/origin-device-registry-irec-local-api/src/device/cron/check-device-state.task.ts
```diff
@@ -20,7 +20,7 @@ export class CheckDeviceStateTask {
         const devices = await this.deviceService.findAll();
         for (const device of devices) {
             const irecDevice = await this.irecService.getDevice(device.ownerId, device.code);
-            if (irecDevice.status !== device.status) {
+            if (irecDevice && irecDevice.status !== device.status) {
```
soanvig, line 23: Ah, strict null checks would be helpful :)

soanvig (approved):

packages/devices/origin-device-registry-irec-local-api/src/device/cron/check-device-state.task.ts
ioncreature: It's just a workaround for IREC mock. It could not contain the required device, so I've added this check to ignore next lines

soanvig: Sure, but if we had strict null checks you would catch this earlier. Nvm though

josipbagaric (approved):

PR #2599, opened 2021-07-06
fix(exchange-io-erc1888): handle blockchain errors

josipbagaric:
`process()` function should not crash the backend due to blockchain errors.


soanvig (approved):
This was crashing whole backend? It should just throw, but not crash the process

PR #2602, opened 2021-07-07
chore(migrations-irec): added irec beneficiaries to seed file

ioncreature:


packages/tools/migrations-irec/config/seed.sql
```diff
@@ -664,4 +664,75 @@ VALUES (
         'd49f7fed-2f79-4d10-8985-41a0c3e9ba03',
         'METER003',
         'Biomass Facility description'
-    );
\ No newline at end of file
+    );
+
+INSERT INTO public."platform_organization" (
```
soanvig: it's minor but maybe https://www.postgresqltutorial.com/postgresql-insert-multiple-rows/
would be more readable?

packages/tools/migrations-irec/config/seed.sql
```diff
@@ -664,4 +664,75 @@ VALUES (
...
+    "createdAt",
+    "updatedAt",
+    id,
+    irecBeneficiaryId,
+    organizationId
+)
+VALUES (
+           '2020-03-30 09:55:25.962333+02',
+           '2020-03-30 09:55:25.962333+02',
+           1,
```
soanvig, line 679: Inserting that manually doesn't increment serial field, if this should be autoincrementing, which will cause conflicts if you want to insert anything later on. is this autoincrement field?

soanvig (changes requested):

packages/tools/migrations-irec/config/seed.sql
ioncreature: fixed by adding `nextval`

packages/tools/migrations-irec/config/seed.sql
ioncreature: fixed

soanvig (approved):

josipbagaric (approved):

PR #2607, opened 2021-07-09
feat: Exchange direction Send

josipbagaric:
Allow sending on exchange to other exchange deposit addresses. 

- Adds a new endpoint `/api/transfer/send`


packages/trade/exchange-io-erc1888/src/withdrawal-processor/withdrawal-processor.service.ts
```diff
@@ -236,20 +227,37 @@ export class WithdrawalProcessorService implements OnModuleInit {
...
+
     private async handleConfirmation(
         transfer: Transfer,
         { transactionHash, logs, blockNumber }: ContractReceipt
     ) {
         const { id } = transfer;
 
-        const logName =
-            transfer.direction === TransferDirection.Withdrawal ? 'TransferSingle' : 'ClaimSingle';
+        const isWithdrawalOrSend = (transfer: Transfer) =>
```
soanvig: minor: you are shadowing a variable here

packages/trade/exchange/migrations/1625819983722-DirectionAndStatusStrings.ts
```diff
@@ -0,0 +1,100 @@
...
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        const oldTransfers = await queryRunner.query(
+            `SELECT id, status, direction FROM "exchange_transfer"`
+        );
+
+        await queryRunner.query(`ALTER TABLE "exchange_transfer" DROP COLUMN "status"`);
+        await queryRunner.query(`ALTER TABLE "exchange_transfer" DROP COLUMN "direction"`);
+        await queryRunner.query(`ALTER TABLE "exchange_transfer" ADD "status" character varying`);
+        await queryRunner.query(
+            `ALTER TABLE "exchange_transfer" ADD "direction" character varying`
```
soanvig, line 48: so you drop this column and add again? To clean up?

josipbagaric: Yeah. I thought about using `ALTER TABLE ... TYPE ... USING <expression>` but I couldn't find the correct expression to use there.

Since the migration is a transaction that is commited only after the `up()` method is fully executed, I think this approach should work fine.

soanvig (approved):

PR #2668, opened 2021-07-21
chore: update device types to I-REC definition

josipbagaric:
Use I-REC fuel/device types.


josipbagaric:
@alexworker23 Updated it to the old way for simplicity purposes, take a look.

alexworker23 (approved):

PR #2669, opened 2021-07-21
Feat/supply page

driamend:


packages/ui/libs/exchange/logic/src/supply/supply.ts
```diff
@@ -0,0 +1,68 @@
...
+          '',
+        facilityName: device?.name,
+        price: `$${price}`,
+        status: matchingSupply.active
+          ? SupplyStatus.Active
+          : SupplyStatus.Paused,
+        toBeCertified: 0,
+        actions,
+      };
+    }) || ([] as ComposedPublicDevice[])
```
alexworker23: Not sure if we want to use this type here. Should return simply `[]`

packages/ui/libs/exchange/logic/src/updateSupplyForm/form.ts
```diff
@@ -0,0 +1,71 @@
+import { useTranslation } from 'react-i18next';
+import * as yup from 'yup';
+import { supplyStatusOptions, SupplyStatus } from './supplyStatusOptions';
+import { TUseUpdateSupplyFormLogic } from './types';
+
+export const UpdateSupplyFormLogic: TUseUpdateSupplyFormLogic = (
```
alexworker23: forgot the `use` to make it custom hook `useUpdate...`

alexworker23 (approved):

PR #2717, opened 2021-08-03
Chore/fixes and improvements

driamend:


packages/ui/libs/user/view/src/pages/LoginPage/LoginPage.effects.ts
```diff
@@ -1,13 +1,27 @@
...
+import { useState } from 'react';
 
 export const useLogInPageEffects = () => {
   const navigate = useNavigate();
   const dispatchModals = useUserModalsDispatch();
 
+  const [formValues, setFormValues] =
+    useState<TUserLoginFormValues>(INITIAL_FORM_VALUES);
+
+  const onWatchHandler = (values: string[]) => {
```
alexworker23, line 19: To be honest, I don't think this is the best way to achieve the desired behaviour of the form, but we'll keep it for now. Probably will dig into validation modes and pick the most suitable for our case.

alexworker23 (approved):

PR #2722, opened 2021-08-05
Chore/fixes and improvements 2

driamend:


packages/ui/libs/ui/core/src/components/layout/PermissionsFeedback/PermissionsFeedback.tsx
```diff
@@ -0,0 +1,43 @@
...
+  accessRules: IPermissionRule[];
+}
+
+export const PermissionsFeedback = memo(({ accessRules }: IProps) => {
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <Typography variant="body1" className="mt-3" gutterBottom>
+        {t('general.requirements.requirementsTitle')}
```
alexworker23: We should not have localization in core components. We should pass them as props e.g. title

packages/ui/libs/ui/core/src/components/layout/Requirements/Requirements.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const classes = useStyles();
+  const { loading, accessRules } = usePermissions(rules);
+
+  if (loading) {
+    return <CircularProgress />;
+  }
+
+  return (
+    <Paper className={classes?.container}>
+      <PermissionsFeedback accessRules={accessRules} />
```
alexworker23: PermissionFeedback can be moved here, without creating additional component. 
I can't see any additional way where re-using PermissionsFeedback will be useful..

packages/ui/libs/ui/utils/src/permissions/hooks/usePermissions.ts
```diff
@@ -1,50 +1,58 @@
...
+} from './../../types';
+
+export const usePermissions = (
+  config = defaultRequirementList
+): IPermissionReturnType => {
+  const { t } = useTranslation();
+
+  const { data: userData, isLoading: userLoading } = useUserControllerMe();
+  const { data: account, isLoading: exchangeAddressLoading } =
+    useAccountControllerGetAccount();
```
alexworker23: origin-ui-utils is a generic package for helpers functions. So we cant store the api calls here, because it will make generic component dependent on the exact implementation of exchange-api and origin-backend. 
You should keep here only the main logic for checking the if the requirements are met. 
All other custom logic should be implemented in a dedicated package.

alexworker23 (changes requested):

packages/ui/libs/certificate/logic/src/blockchain-inbox/permissions.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  usePermissionsLogic,
+  TUsePermissions,
+  IPermissionReturnType,
+  RequirementList,
+  Requirement,
+} from '@energyweb/origin-ui-certificate-logic';
```
alexworker23: You shouldn't use such import paths inside the package itself, it will cause circular dependency error. 
Use relative path `../../someFile`

alexworker23 (approved):

PR #2724, opened 2021-08-06
chore: simplify CreateDemand implementation

soanvig:
Either way it's breaking change, so we must decide whether to include end date or not (the default value on backend)


packages/ui/exchange-ui-core/src/components/modal/DemandUpdateModal.tsx
```diff
@@ -102,8 +102,7 @@ export function DemandUpdateModal(props: IProps) {
                 start: moment(values.start).toISOString(),
                 end: moment(values.end).toISOString(),
                 product: values.product,
-                boundToGenerationTime: false,
-                excludeEnd: false
```
soanvig, line 106: I set default excludeEnd to `true` on backend

packages/ui/exchange-ui-core/src/pages/ViewMarketPage/hooks/useBidHandler.ts
```diff
@@ -68,8 +68,7 @@ export async function bidHandler({
                         ? undefined
                         : values.location?.map((l) => `${country};${l}`)
                 },
-                boundToGenerationTime: false,
-                excludeEnd: false
```
soanvig, line 72: I set default excludeEnd to `true` on backend

packages/ui/exchange-ui-core/src/utils/exchange/math.ts
```diff
@@ -23,15 +23,13 @@ export async function calculateTotalVolume(
...
-                parsedVolume
-            ).toString(),
+            volumePerPeriod:
+                EnergyFormatter.getBaseValueFromValueInDisplayUnit(parsedVolume).toString(),
             periodTimeFrame: period,
             start: start.toISOString(),
             end: end.toISOString(),
             product: {},
-            boundToGenerationTime: false,
-            excludeEnd: false
```
soanvig, line 34: I set default excludeEnd to `true` on backend

kosecki123:
What's the reason for this change and what does this simplify? It was actually explicitly added here https://github.com/energywebfoundation/origin/pull/1342

soanvig:
> What's the reason for this change and what does this simplify? It was actually explicitly added here #1342

I know, let's wait for @JosephBagaric with specific explanation, because I'm not entirely sure. I just got info what should be done :)

kosecki123 (approved):

PR #2743, opened 2021-08-10
feat(issuer): Add RegistryExtended contract

josipbagaric:
Adds a new smart contract RegistryExtended which currently contains all the `BatchMultiple` operations.


ioncreature (approved):

packages/traceability/issuer-api/src/pods/certificate/dto/batch-certificate-transfer.dto.ts
```diff
@@ -0,0 +1,26 @@
...
+    implements Omit<CertificateBatchOperations.BatchCertificateTransfer, 'amount'>
+{
+    @ApiProperty({ type: String })
+    @IsEthereumAddress()
+    to: string;
+
+    @ApiPropertyOptional({ type: String })
+    @IsOptional()
+    @IsEthereumAddress()
+    from?: string;
```
soanvig, line 20: Why is this optional?

packages/traceability/issuer-api/src/pods/certificate/dto/batch-certificate-claim.dto.ts
```diff
@@ -0,0 +1,18 @@
...
+    extends OmitType(BatchCertificateTransferDTO, ['to'] as const)
+    implements Omit<CertificateBatchOperations.BatchCertificateClaim, 'amount'>
+{
+    @ApiProperty({ type: Object })
+    @IsObject()
+    claimData: IClaimData;
+
+    @ApiPropertyOptional({ type: String })
+    @IsEthereumAddress()
+    to?: string;
```
soanvig, line 17: Why is it optional?

packages/traceability/issuer-api/src/pods/certificate/handlers/batch-claim-certificates.handler.ts
```diff
@@ -20,59 +19,43 @@ export class BatchClaimCertificatesHandler
...
-    }: BatchClaimCertificatesCommand): Promise<ISuccessResponse> {
+    async execute({ claims }: BatchClaimCertificatesCommand): Promise<ISuccessResponse> {
         const blockchainProperties = await this.blockchainPropertiesService.get();
 
-        const certificatesToClaim = await this.repository.findByIds(
-            certificateAmounts.map((cert) => cert.id),
-            { relations: ['blockchain'] }
-        );
+        for (const { id, amount, from } of claims) {
+            if (!amount) {
```
soanvig, line 26: I think this will not work. If amount is a string, then `!'0'` returns false

packages/traceability/issuer-api/src/pods/certificate/handlers/batch-transfer-certificates.handler.ts
```diff
@@ -20,50 +19,34 @@ export class BatchTransferCertificatesHandler
...
-    }: BatchTransferCertificatesCommand): Promise<ISuccessResponse> {
+    async execute({ transfers }: BatchTransferCertificatesCommand): Promise<ISuccessResponse> {
         const blockchainProperties = await this.blockchainPropertiesService.get();
 
-        const certificatesToTransfer = await this.repository.findByIds(
-            certificateAmounts.map((cert) => cert.id),
-            { relations: ['blockchain'] }
-        );
+        for (const { id, amount, from } of transfers) {
+            if (!amount) {
```
soanvig, line 26: the same

soanvig (changes requested):

packages/traceability/issuer-api/src/pods/certificate/dto/batch-certificate-transfer.dto.ts
josipbagaric: If it's not defined, then we assume that the transfer is `from` the caller, `to` someone else.

packages/traceability/issuer-api/src/pods/certificate/dto/batch-certificate-claim.dto.ts
josipbagaric: `to` is only used if you want to claim `to` someone else's behalf.
If it remains undefined, it is implied that you are claiming to you own address.

packages/traceability/issuer-api/src/pods/certificate/handlers/batch-claim-certificates.handler.ts
josipbagaric: Maybe it's not that explicit, but the amount is of type `BigNumber`. 
`!BigNumber.from('0')` is still false.

soanvig: So it's not working as expected, right? It will always be false, therefore will never ignore this claim and always process it

josipbagaric: Ok to be more precise, `amount` can be either `BigNumber` or `undefined`. If it's undefined, `continue` is triggered.

soanvig: Oh, i get it, i thought it's for ignoring claims for 0 volume :)

soanvig (approved):

PR #2747, opened 2021-08-12
feat(issuer-api): automatically run contracts deployment

soanvig:
I also added start:dev command


soanvig:
i need to fix some things still

soanvig:
I pause this PR because there are a lot of problems. I will fix this in feature

PR #2763, opened 2021-08-16
Feat/issuer tx tracking

soanvig:
1. added new entity that tracks transactions happening on blockchain, add query (`GetCertificatesWithLogsQuery`)
2. (BREAKING CHANGE) renamed AppModule to IssuerModule -> made IssuerModule dynamic and configurable (so transaction tracking can be disabled). User has to use IssuerModule directly now. IssuerModule has been cleaned off any "root" dependencies (ValidationPipe and ScheduleModule)


packages/traceability/issuer-api/src/app.module.ts
```diff
@@ -1,45 +0,0 @@
...
-            isGlobal: true
-        }),
-        ScheduleModule.forRoot(),
-        CertificateModule,
-        BlockchainPropertiesModule,
-        CertificationRequestModule
-    ],
-    providers
-})
-export class AppModule {}
```
josipbagaric, line 45: This would trigger a breaking change. But since we already have a bunch of breaking changes, it's fine.

packages/traceability/issuer-api/src/pods/certificate/certificate.entity.ts
```diff
@@ -61,4 +60,10 @@ export class Certificate extends ExtendedBaseEntity {
     @Column()
     @IsBoolean()
     issuedPrivately: boolean;
+
+    @CreateDateColumn({ type: 'timestamptz' })
+    createdAt: Date;
+
+    @UpdateDateColumn({ type: 'timestamptz' })
+    updatedAt: Date;
```
josipbagaric, line 68: Any reason for this change?

josipbagaric (commented):
Just took a short look to see where this is heading.

Are you sure you based the branch on the latest master?

I see that some changes here no longer exist in master, such as `TOTAL_AMOUNT`.

packages/traceability/issuer-api/src/app.module.ts
soanvig: Yep, it's marked in the commit that this is breaking change, I corrected all packages as well (I hope SO)

packages/traceability/issuer-api/src/pods/certificate/certificate.entity.ts
soanvig: ExtendedBaseEntity which implements createdAt and updatedAt is ActiveRecord in fact, and either I create `ICertificate` or something like that, or make this entity not ActiveRecord. We don't make use of this entity being the ActiveRecord, we use repository pattern everywhere.

The problem with ActiveRecord is that it cannot be easily created as an object (`{ ... }`), it also sneaks methods like save, create, update which are dirty (since certificate entites are returned from issuer-api to other services)

josipbagaric: Cool, I remember now. I'm fine with this.

soanvig:
@JosephBagaric  
> Are you sure you based the branch on the latest master?

I rebased it, because I had a conflict. I had some conflict in a file where`TOTAL_AMOUNT` is so that may be a reason. I will remove this.

packages/traceability/issuer-api/src/pods/certificate/transaction-log.entity.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const TRANSACTION_LOG_TABLE_NAME = 'transaction_log';
+
+@Entity({ name: TRANSACTION_LOG_TABLE_NAME })
+export class TransactionLog {
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Index()
+    @Column({ type: 'int4' })
```
josipbagaric: Use `bigint` instead of `int4`, since the certificate ID can be up to `2^256 - 1`.

josipbagaric: This also needs to be changed in other parts of the codebase, but let's leave it for another time.

josipbagaric: We should probably also add a check `Min(1)`.

soanvig:
This should not be included in any stable release, that we are going to do, since there is this issuer-api change, which is risky to be honest (because of these changed modules dependencies), and some things may NOT work if they are not tested e2e (handlers or something)

packages/traceability/issuer-api/src/pods/certificate/transaction-log.entity.ts
soanvig: What's that check? Is it from TypeOrm? I cannot find this

josipbagaric: `class-validator` package

josipbagaric: Like in https://github.com/energywebfoundation/origin/blob/0cdd344dbc718d5048457c95d4d2dada4d45ad96/packages/traceability/issuer-api/src/pods/certificate/certificate.entity.ts#L13

soanvig: But why to use class-validator here? It's not validated anywhere (transaction log cannot be created as a input from API or something where it would be validated by Nest), and the object before creation is not validated by class-validator manually anywhere.

Afaik adding this decorator changes nothing, if we don't explicitly or implicitly validate this with class-validator functions.

Could you shed more light on this?

josipbagaric: If we add class-validator in the entity, it will be validated on `repository.save()` if I remember correctly. This should be handled by TypeORM. 

Indeed, we'll almost 100% always get a valid certificate ID from an event, but still it's an additional layer of validation.

I'm also fine if you choose to leave this out.

josipbagaric: https://orkhan.gitbook.io/typeorm/docs/validation

josipbagaric: Seems like you might need to manually call `validate()` before `.save()`.

soanvig: I can do this I think

soanvig: Added! :)

packages/traceability/issuer-api/src/pods/certificate/transaction-log.service.ts
```diff
@@ -0,0 +1,169 @@
...
+            certificateId,
+            transactionHash,
+            transactionType,
+            transactionTimestamp
+        });
+
+        const errors = await validate(entity);
+
+        if (errors.length > 0) {
+            throw new Error(`Transaction save failed: ${errors.join(', ')}`);
```
josipbagaric: We can also replace `Error` with `UnprocessableEntityException()` from NestJS, to be more explicit. Not a big deal though.

josipbagaric (approved):

josipbagaric:
> This should not be included in any stable release, that we are going to do, since there is this issuer-api change, which is risky to be honest (because of these changed modules dependencies), and some things may NOT work if they are not tested e2e (handlers or something)

We'll be releasing a stable version soon. So you're suggesting not to include this?

soanvig:
> We'll be releasing a stable version soon. So you're suggesting not to include this?

If it will be thouroughly tested, then why not. If we are going to believe, that everything works, then it's better to not include this in stable release, and test this over time. I had to touch multiple applications (without extensive knowledge about them), that's why I am cautious

josipbagaric:
> > We'll be releasing a stable version soon. So you're suggesting not to include this?
> 
> 
> 
> If it will be thouroughly tested, then why not. If we are going to believe, that everything works, then it's better to not include this in stable release, and test this over time. I had to touch multiple applications (without extensive knowledge about them), that's why I am cautious

Well, we have decent test coverage so we can assume it's fine when the tests pass.

If you're still sceptic, you can run the test suite on Volta instead of Ganache and see if anything breaks.

ioncreature (approved):

soanvig:
@JosephBagaric  unfortunately I had to remove validation - because certificateId is a string now (BigInt) it's not that easy to validate that using class-validator without class-transfomer usage, and that would require installing dependency.

While validating certificateId is not that important IMHO since if it will be like `-1` or something, then it won't be find by querying by certificate id anyway.

josipbagaric:
Cool. Let's wait with merging this PR for now.

josipbagaric (approved):

PR #2764, opened 2021-08-17
chore(issuer): Don't transfer when claiming if same sender/receiver

josipbagaric:


packages/traceability/issuer/contracts/RegistryExtended.sol
```diff
@@ -120,7 +120,9 @@ contract RegistryExtended is Registry {
 			Certificate memory cert = certificateStorage[_ids[i]];
 			topics[i] = cert.topic;
 
-            safeTransferFrom(_from[i], _to[i], _ids[i], _values[i], _data[i]);
+			if (_from[i] != _to[i]) {
+            	_safeTransferFrom(_from[i], _to[i], _ids[i], _values[i], _data[i]);
+			}
 
 			_burn(_to[i], _ids[i], _values[i]);
```
ioncreature, line 127: Seems like we have to move `_burn(_to[i], _ids[i], _values[i]);` into `if` statement

ioncreature: And use spaces instead of tabs please 😄

josipbagaric: The `burn` method should stay outside the `if` clause, because we always burn the tokens - regardless whether the `from` and `to` addresses are the same.

josipbagaric: We just skip transferring the tokens if the `to` and `from` are the same address, because a transfer is unnecessary.

ioncreature: Ok, seems reasonable

ioncreature (approved):

PR #2768, opened 2021-08-19
chore(origin-device-registry-api): sanitize call filtered all data

ioncreature:


packages/devices/origin-device-registry-api/src/device-registry/dto/new-device.dto.ts
```diff
@@ -35,6 +35,6 @@ export class NewDeviceDTO {
     imageIds?: string[];
 
     public static sanitize(device: NewDeviceDTO): NewDeviceDTO {
-        return plainToClass(NewDeviceDTO, device, { excludeExtraneousValues: true });
+        return plainToClass(NewDeviceDTO, device);
```
ioncreature, line 38: @JosephBagaric not sure it is super safe, but typeorm does the validation anyway. And this `{ excludeExtraneousValues: true }` removed all the data from `device`.

josipbagaric: Yeah this is a known issue. Happened to me more than once. 🤷

josipbagaric (approved):

PR #2842, opened 2021-09-06
fix(origin-organization-irec-api): suppress error on list irec accounts for registrant

ioncreature:


packages/organizations/origin-organization-irec-api/src/connection/handlers/get-accounts.handler.ts
```diff
@@ -21,6 +21,12 @@ export class GetAccountsHandler implements ICommandHandler<GetAccountsCommand> {
     ) {}
 
     async execute({ owner }: GetAccountsCommand): Promise<AccountDTO[]> {
-        return this.irecService.getAccountInfo(owner);
+        try {
+            return await this.irecService.getAccountInfo(owner);
+        } catch (e) {
+            // Registrant account does not have permissions to get list of accounts
+            // so we return empty list
+            return [];
```
josipbagaric: Shouldn't we throw a 403 instead?

josipbagaric: And return the description message that we got from I-REC API.

ioncreature: I think no, we should not. It is just a list of accounts. And since the user does not have accounts empty list is better solution

josipbagaric: Let's at least add some logger log that says why we return an empty list.

josipbagaric: Something like "user with ID requested accounts but it has none".

josipbagaric: Will the I-REC API ever throw any other error besides this one when we do `getAccountInfo`?

josipbagaric: Would be better to do in `catch(e)` something like 
```js
if (e.status === HttpStatus.FORBIDDEN) {
    logger.log("....");
    return [];
}

throw e;
``` 

This way we would make sure that other errors are thrown and we can identify them, such as connection to I-REC API being broken etc.

ioncreature: Done. Had to update irec client library for this

josipbagaric (approved):

josipbagaric:
👌

PR #2844, opened 2021-09-07
Origin-UI version 2

alexworker23:
# Origin UI version 2

## Extended and more generic core libraries
- `origin-ui-assets`: Stores all icons and images required for building Origin UI
- `origin-ui-blockchain`: Library with basic helper functions, providers and HOC's for web3 interactions
- `origin-ui-core`: Completely new library containing only Generic React components
- `origin-ui-localization`: Localization library for Origin UI
- `origin-ui-shared-state`: Helper functions library used for managing localStorage interactions
- `origin-ui-theme`: Package with pre-configured Origin MUI theme and some coloring helper functions
- `origin-ui-utils`: General utils library for formatting data

## New UI apps structure
- Each part of the UI (Device, Certificate, Exchange, Organization, User) is designed to run autonomously without depending on other UI apps, but only dependent to core ui packages. 
- Each UI app is split into 3 layers/packages: data, logic, view. 

## Tech stack changes
- Using `nx` monorepo tooling for managing all UI packages
- MUI packages v5
- Instead of `redux` and `redux-saga` now using `react-query`
- Replaced usage of openapi auto-generated clients by using `orval` generated `react-query` api-client libs
- `react-hook-form` instead of `formik`

## Environment variables handling
Currently env variables on UI are split into 2 sources: 
1. Secret credentials are passed on the build time via `env.sh` & `.env.local`.
2. Public variables are passed at the runtime through `window` object in index.html and handled via `dev-env.sh` & `prod-start.sh`

## View changes

All Devices

Screenshot 2021-09-08 at 12 18 03

My Devices

Screenshot 2021-09-08 at 12 18 14

Request certificates

Screenshot 2021-09-08 at 12 18 29

Device Detail View

Screenshot 2021-09-08 at 12 19 28

Register Device

Screenshot 2021-09-08 at 12 18 47

Theme Switcher

Screenshot 2021-09-08 at 12 21 02


README.md
```diff
@@ -126,6 +137,8 @@ rush test:e2e
 
 ## Run demo
 
+=======
```
josipbagaric: This is a remainder from some merge?

alexworker23: Yes, will remove

packages/api-clients/react-query/origin-backend/src/response-type.ts
```diff
@@ -0,0 +1,15 @@
...
+
+    // @ts-ignore
+    promise.cancel = () => {
+        source.cancel('Query was cancelled by React Query');
+    };
+
+    return promise;
+};
+
+export default getWithResponseType;
```
josipbagaric, line 15: Noticed we have this file in every client folder. Is there a better way?

packages/api-clients/react-query/origin-organization-irec-api/package.json
```diff
@@ -0,0 +1,57 @@
+{
+    "name": "@energyweb/origin-organization-irec-api-react-query-client",
```
josipbagaric, line 2: Should this client libs be available on NPM?

If not, we should put the private flag in their package.json.

packages/api-clients/react-query/origin-backend/src/response-type.ts
alexworker23: While developing we also had this concern, but at that time didn't find a better way to handle this.
But I will try to look for a better solution now

packages/api-clients/react-query/origin-organization-irec-api/package.json
alexworker23: Why not? Previously we were releasing it on npm

josipbagaric: We did because we needed them for CD. Do we still need them for deployments? I think now we build all docker images locally and then deploy.

alexworker23: But what about other users setting up the origin-ui? 
It is required by Certificate, Organization part of the UI and the root itself

packages/trade/exchange-client/tsconfig.json
```diff
@@ -4,7 +4,7 @@
         "outDir": "./dist/js",
         "emitDecoratorMetadata": true,
         "experimentalDecorators": true,
-        "target": "ES2020",
+        "target": "es2015",
         "types": ["node", "mocha"]
     },
     "include": ["src/**/*"]
```
josipbagaric, line 10: Hmm do we still even need this package exchange-client?

josipbagaric: Since we have react query clients.

alexworker23: Not sure, maybe we can keep them alongside react-query clients in case anyone wants to use the regular ones.
But the main thing is that we have to change `target` in all packages from `es2020`  to `es2015`, because it makes no sense in keeping it `es2020` and also can cause bugs and problems.

packages/ui/env.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+REQUIRED_VARIABLES=(
+  EXCHANGE_WALLET_PUB
+  GOOGLE_MAPS_API_KEY
+  ISSUER_ID
+  SMART_METER_ID
+  REGISTRATION_MESSAGE_TO_SIGN
+)
```
josipbagaric, line 9: There are some references in the code to variables:

NX_GOOGLE_MAPS_API_KEY
NX_EXCHANGE_WALLET_PUB
NX_REGISTRATION_MESSAGE_TO_SIGN
NX_SMART_METER_ID
BACKEND_URL
BLOCKCHAIN_EXPLORER_URL
SUPPORTED_NETWORK_IDS

Should these also be included here?

packages/ui/apps/origin-ui/src/hooks/useInterceptors.ts
```diff
@@ -0,0 +1,26 @@
+import { getAuthenticationToken } from '@energyweb/origin-ui-shared-state';
+import axios from 'axios';
+
+declare global {
+  interface Window {
+    config: {
+      BACKEND_URL: string;
+      BLOCKCHAIN_EXPLORER_URL: string;
+      SUPPORTED_NETWORK_IDS: string;
```
josipbagaric, line 9: Do we need to add other variables here such as:
NX_GOOGLE_MAPS_API_KEY
NX_EXCHANGE_WALLET_PUB
NX_REGISTRATION_MESSAGE_TO_SIGN
NX_SMART_METER_ID

packages/ui/env.sh
alexworker23: As I have mentioned in PR description our UI env variables are now split between 2 sources. 
1. `env.sh` handles the secrets part which is added on build time and also adds `NX_` prefixes.

2. `dev-env.sh` and `prod-start.sh` are responsible for adding 3 other variables:
BACKEND_URL
BLOCKCHAIN_EXPLORER_URL
SUPPORTED_NETWORK_IDS
They can be adjusted on the runtime and they are publicly available through `window.config` since there is no reason in hiding them, but we can use this and change them at runtime.

packages/ui/apps/origin-ui/src/hooks/useInterceptors.ts
alexworker23: No, because then they will be publicly available and we don't want this to happen

packages/api-clients/react-query/origin-backend/src/response-type.ts
alexworker23: To be honest cannot find a better way of storing this file elsewhere

josipbagaric (approved):

Malak67 (approved):

PR #2845, opened 2021-09-08
feat(issuer-irec-api): added isDeviceImported field to list of certificates to import

ioncreature:


packages/traceability/issuer-irec-api/src/pods/certificate/handler/get-irec-certificates-to-import.handler.ts
```diff
@@ -17,20 +18,29 @@ export class GetIrecCertificatesToImportHandler
...
 
-        return irecCertificates.filter((issue) => {
-            return !certificationRequests.some((cr) => issue.asset === cr.irecAssetId);
-        });
+        return irecCertificates
+            .filter((issue) => {
+                return !certificationRequests.some((cr) => issue.asset === cr.irecAssetId);
+            })
+            .map((issue) => {
+                const irecDevice = irecDevices.find((d) => d.code === issue.device.code);
```
soanvig: if you use `some` instead of `find` you receive boolean, therefore no need to convert to Boolean (with !! which I don't like) later on

soanvig (approved):

packages/traceability/issuer-irec-api/src/pods/certificate/handler/get-irec-certificates-to-import.handler.ts
ioncreature: fixed

soanvig (approved):

PR #2846, opened 2021-09-09
feat/confirm email page

alexworker23:
In this PR:
- Added confirm email page
- Adjusted the Re-send email confirmation localization and submitHandler
- Add custom hook to utils package for receiving query params from location
- Added "Email confirmed" field to User Profile page
- For `backend-irec-app` adjusted the urls which are sent to users via email on organization invitations and email confirmations

### DIFF
--- a/packages/ui/libs/ui/localization/src/languages/en.json
+++ b/packages/ui/libs/ui/localization/src/languages/en.json
@@ -822,6 +822,12 @@
           "loginError": "Could not log in with supplied credentials"
         }
       },
+      "confirmEmail": {
+        "alreadyConfirmed": "Email has been already confirmed",
+        "success": "Email successfully confirmed",
+        "expired": "Confirmation link expired",
+        "error": "Error while confirming email"
+      },
       "profile": {
         "userInfoTitle": "User information",
         "changeEmailTitle": "Change email",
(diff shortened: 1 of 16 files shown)


packages/ui/libs/ui/localization/src/languages/en.json
```diff
@@ -852,6 +861,8 @@
           "blockchainHowTo": "Select the right blockchain account in your MetaMask browser wallet and click “Connect Blockchain Address” to connect. Visit metamask.io to set up your own blockchain wallet."
         },
         "notifications": {
+          "confirmationEmailResentSuccess": "Confirmation email re-sent",
```
Malak67, line 864: I think we shoud write it resent without the '-'

packages/ui/libs/ui/localization/src/languages/en.json
```diff
@@ -852,6 +861,8 @@
...
+          "confirmationEmailResentError": "Error while trying to re-send confirmation email",
```
Malak67, line 865: Same here. Just 'resend'. I would also change this in the backend (endpoint name)

packages/ui/libs/ui/localization/src/languages/en.json
alexworker23: resent - is a word with different meaning, which can cause confusion

packages/ui/libs/ui/localization/src/languages/en.json
alexworker23: We can change it here, but since we have a `re-sent` I would suggest to keep it here as well to be aligned on the naming

Malak67 (approved):

packages/ui/libs/ui/localization/src/languages/en.json
Malak67: still sounds a bit weird. But then you would need to change it everywhere in the backend. So you can leave it.

PR #2850, opened 2021-09-10
fix/master deploy

alexworker23:
- Removed not needed heroku deployments url's
- Removed no longer used packages from rush configuration
- Added build time env variables to CI workflows
- Remove ui-v2 build/deploy scripts from ui Makefile


rush.json
```diff
@@ -440,10 +428,6 @@
             "packageName": "@energyweb/origin-backend-core",
             "projectFolder": "packages/origin-backend-core"
         },
-        {
-            "packageName": "@energyweb/migrations",
-            "projectFolder": "packages/tools/migrations"
-        },
         {
             "packageName": "@energyweb/migrations-irec",
             "projectFolder": "packages/tools/migrations-irec"
```
josipbagaric, line 433: What's the reason for removing them from the project index?

alexworker23: So that we won't build them or run any other commands in them

josipbagaric: Hmm, I guess that's fine for now until we move them elsewhere.

josipbagaric (approved):

PR #2851, opened 2021-09-10
fix/remove docker build

alexworker23:


josipbagaric:
Why are these being removed?

alexworker23:
> Why are these being removed?

Because on master deployment when running bulk command `yarn build:containers:canary ` we receive an error:
`ERROR: The project "@energyweb/origin-backend-app" does not appear in the list of "deploymentProjectNames" from deploy.json. `

josipbagaric (approved):

PR #2893, opened 2021-09-14
feat(origin-device-registry-local-irec-api): capacity is string now

ioncreature:


packages/devices/origin-device-registry-irec-local-api/src/device/device.service.ts
```diff
@@ -66,7 +66,7 @@ export class DeviceService {
 
         const irecDevice = await this.irecService.createDevice(user, {
             ...deviceData,
-            capacity: deviceData.capacity / 1e6,
+            capacity: String(Number(deviceData.capacity) / 1e6),
```
josipbagaric: `Number(deviceData.capacity)` will overflow if the number is too big.

We should use BigNumber here.
```js
BigNumber.from(deviceData.capacity).div(1e6).toString()
```

josipbagaric (changes requested):

packages/devices/origin-device-registry-irec-local-api/migrations/1631576581265-CapacityString.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class CapacityString1631576581265 implements MigrationInterface {
+    name = 'CapacityString1631576581265';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            `ALTER TABLE "irec_device_registry_device" ALTER COLUMN "capacity" TYPE character varying`
+        );
+    }
```
josipbagaric, line 10: Please create a migration for existing device capacities, it should be simple.

packages/devices/origin-device-registry-irec-local-api/src/device/device.service.ts
ioncreature: Number.MAX_SAFE_INTEGER is 9007199254740991, I think it is enough

packages/devices/origin-device-registry-irec-local-api/migrations/1631576581265-CapacityString.ts
ioncreature: Altering types do not delete data, it changes int to string preserving all the data we had before

packages/devices/origin-device-registry-irec-local-api/src/device/device.service.ts
ioncreature: fixed

alexworker23 (approved):

josipbagaric (approved):

PR #2905, opened 2021-09-17
[issuer-api]: increase transaction log test consistency

soanvig:
I could not find the issue why it sometimes finds 3 certificates for given device-id. It fails randomly, and I couldn't reproduce this locally.

But my guess is that's some leftover from other test (like some transaction that was hanging on blockchain and it gets processed, and creates entries in database).

Because all tests in issuer-api where using the same device name, now tests in transaction logs are using different names (for each single test), so this should fix the issue while preserving tests usable.

### DIFF
--- a/packages/traceability/issuer-api/test/transaction-logs.e2e-spec.ts
+++ b/packages/traceability/issuer-api/test/transaction-logs.e2e-spec.ts
@@ -28,7 +28,6 @@ const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
 
 const certificateTestData = {
     to: deviceManager.address,
-    deviceId: 'ABC-123',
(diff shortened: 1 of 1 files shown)


soanvig:
Do you know how can I restart deepscan?

josipbagaric:
> Do you know how can I restart deepscan?

Not sure, but you can try rebasing the commit, or creating an empty commit with `git commit --allow-empty`. This should re-trigger the deepscan.

soanvig:
> Not sure, but you can try rebasing the commit, or creating an empty commit with `git commit --allow-empty`. This should re-trigger the deepscan.

Oh yeah that helped

soanvig:
Why the heck it shows 39 issues from not my changes

josipbagaric (approved):

JonaszJestem (approved):

PR #2934, opened 2021-09-24
feat/timezones

alexworker23:
In this PR: 

- Added an array of Time Zones to `utils-general` package

- Adjusted 2 forms:
1. Register Device 
Removed country field, now consuming country from the `/api/configuration`
Added timezone field, which is optional and appears only in case the country of the platform contains more than 1 timezone
2. Import Device
Made timezone as an optional selector, which appears only in case the country of the platform contains more than 1 timezone.
Made regions and subregions selectors instead of regular input.

- Removed unnecessary validation from Generic Form which was related to it being a part of MultiStepForm


.env.example
```diff
@@ -21,6 +21,9 @@ INFLUXDB_ADMIN_PASSWORD=admin
 INFLUXDB_USER=api
 INFLUXDB_USER_PASSWORD=secretpassword
 
+# Country of platform deployment and usage. Should be specified following ISO Alpha-2 format
+PLATFORM_COUNTRY_CODE=TH
```
josipbagaric: Don't we have this already defined in /api/configuration?

alexworker23: Good point. Yes, we have it.
I will have to force push the branch in order to remove the breaking_change commit

ioncreature (approved):

PR #2969, opened 2021-10-04
chore: Don't wait for transactions

josipbagaric:
Currently, when we issue/transfer/claim or batch issue/transfer/claim through the Issuer API, we wait for the transaction to be mined before we return a response to the API caller.

This can sometimes lead to failed calls and unresponsiveness if the mining of a certain transaction would take too long.

Solution

1. When the user calls the /transfer /issue /claim endpoints, trigger the transaction and return a txHash.2. Create an on-chain listener for transactions and only update the state once a transaction has been mined

Implications to API users

Not much would have to be changed on the Issuer API itself, but the bigger implications would be to the Issuer API users.The users would have to take the transaction hash returned by the Issuer API and create a listener service that would notify the user that the tx has been mined.

## Checklist
- [x] Return transaction on Certificate and CertificationRequest
- [x] Write proper migrations
- [x] Add CQRS events for mined transactions, along with transaction hashes


packages/traceability/issuer-api/migrations/1633507266818-StoreTxHashInsteadBlock.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export class StoreTxHashInsteadBlock1633507266818 implements MigrationInterface {
+    name = 'StoreTxHashInsteadBlock1633507266818';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            `ALTER TABLE "issuer_certificate" ADD "creationTransactionHash" character varying`
+        );
+        const existingCertificates = await queryRunner.query(`SELECT id FROM "issuer_certificate"`);
+        console.log({
```
soanvig: here

packages/traceability/issuer-api/migrations/1633507266818-StoreTxHashInsteadBlock.ts
```diff
@@ -0,0 +1,105 @@
...
+import { BlockchainProperties } from '../src/pods/blockchain/blockchain-properties.entity';
+
+export class StoreTxHashInsteadBlock1633507266818 implements MigrationInterface {
+    name = 'StoreTxHashInsteadBlock1633507266818';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            `ALTER TABLE "issuer_certificate" ADD "creationTransactionHash" character varying`
+        );
+        const existingCertificates = await queryRunner.query(`SELECT id FROM "issuer_certificate"`);
```
soanvig, line 15: This can get potentially ugly on RAM usage, but it's fine besides that.

packages/traceability/issuer-api/migrations/1633507266818-StoreTxHashInsteadBlock.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        if (existingCertificates.length > 0) {
+            const blockchainProperties: BlockchainProperties = await queryRunner.query(
+                `SELECT * FROM "issuer_blockchain_properties"`
+            );
+            const wrapped: IBlockchainProperties =
+                this.getBlockchainProperties(blockchainProperties);
+
+            const syncedCertificates = await Promise.all(
+                existingCertificates.map((cert: any) => new Certificate(cert.id, wrapped).sync)
```
soanvig, line 25: How exausting this may be? For one case we can have around 30k certificates before we even execute this.

packages/traceability/issuer-api/src/pods/certificate/handlers/certificate-created.handler.ts
```diff
@@ -63,21 +69,47 @@ export class CertificateCreatedHandler implements IEventHandler<CertificateCreat
...
-            latestCommitment
-        });
-
-        this.eventBus.publish(new CertificatePersistedEvent(certificate.id));
+        const txHash = cert.creationTransactionHash.toLowerCase();
+        const unminedCommitment = await this.checkCommitment(txHash);
+
+        const queryRunner = this.connection.createQueryRunner();
+        await queryRunner.connect();
+        await queryRunner.startTransaction();
```
soanvig, line 82: I don't see transaction `commit`. Or does this query runner doesn't work like that?

soanvig (commented):
Is polling for transactions neccessary? It will get super tedious and not-performant.
Can't some event be emitted about finished transaction (whatever transaction kind it was) besides that?

josipbagaric:
> Is polling for transactions neccessary? It will get super tedious and not-performant. Can't some event be emitted about finished transaction (whatever transaction kind it was) besides that?

An event is emitted `CertificatePersistedEvent`. We could implement some event listener, but then it isn't obvious to me that it would be much more performant on Ganache. Also, then it wouldn't be a proper e2e test really.

packages/traceability/issuer-api/src/pods/certificate/handlers/certificate-created.handler.ts
josipbagaric: You are right, there should be a commit. I guess it auto-assumes on release that a commit was wanted and does it automatically.

Will add it explicitly also too.

soanvig:
@JosephBagaric is `CertificatePersistedEvent` for all transaction types: issuance, transfer and transaction? Can it be used instead of polling for transaction hash to check whether transaction was  mined?

I plan to use polling, but only for application startup to check any leftovers in database (for example after application crash), but otherwise I would like to rely on event.

josipbagaric:
> @JosephBagaric is `CertificatePersistedEvent` for all transaction types: issuance, transfer and transaction? Can it be used instead of polling for transaction hash to check whether transaction was mined?
> 
> I plan to use polling, but only for application startup to check any leftovers in database (for example after application crash), but otherwise I would like to rely on event.

Actually no, only for creation. For others you would need to poll the tx.

packages/traceability/issuer-api/migrations/1633507266818-StoreTxHashInsteadBlock.ts
josipbagaric: Dunno, would need to be stress-tested. But since it's a _read_ and not _write_, the RPC node should handle it fine.

josipbagaric: Depends on the RPC node entirely, but you're right, it might trigger some DDoS protection.

josipbagaric:
@soanvig take a look

soanvig (approved):

PR #2971, opened 2021-10-06
chore: Unify ormconfigs

josipbagaric:


josipbagaric:
@artursudnik Take a look if this will work?

artursudnik (approved):
Looks good and what we need!

PR #3018, opened 2021-10-20
feat/UI exchange export

alexworker23:
- Added Export action for Exchange Inbox

Screenshot 2021-10-20 at 11 19 17

- UI dev-server now has Hot Module Replacement enabled, and root (`origin-ui`) is always build to prod on `rush build`


packages/ui/package.json
```diff
@@ -19,7 +19,7 @@
     "build:device-ui": "nx run device-data:build && nx run device-logic:build && nx run device-view:build",
     "build:exchange-ui": "nx run exchange-data:build && nx run exchange-logic:build && nx run exchange-view:build",
     "build:all-packages": "yarn build:ui-utils && yarn build:ui-localization && yarn build:ui-shared-state && yarn build:ui-theme && yarn build:ui-assets && yarn build:ui-core && yarn build:ui-blockchain && yarn build:user-ui && yarn build:organization-ui && yarn build:certificate-ui && yarn build:device-ui && yarn build:exchange-ui",
-    "build:origin-ui": "cp apps/origin-ui/src/index.html.template apps/origin-ui/src/index.html && nx build",
+    "build:origin-ui": "cp apps/origin-ui/src/index.html.template apps/origin-ui/src/index.html && nx build --prod",
```
josipbagaric: Will this increase the build time by much?

josipbagaric (approved):

packages/ui/package.json
alexworker23: no

alexworker23: @JosephBagaric increased by 40s

josipbagaric: Is it necessary to run the full prod build on every build? Maybe just on master build. 40s is quite longer.

alexworker23: Agree, will adjust this in a separate commit

.github/workflows/deploy-master.yml
```diff
@@ -70,6 +70,10 @@ jobs:
     - name: Build application
       run: |
         node common/scripts/install-run-rush.js rebuild -v
+      
+    - name: Build UI for production
```
alexworker23, line 74: @JosephBagaric because there is no way to extract only origin-ui from building on `rebuild` command I have to first build it regular version and afterward build it for production. This is additional 80s, but this will happen only on master and release branches. Other solution is to override default `rush rebuild` command, which I personally don't like.

josipbagaric: That's fine.

josipbagaric (approved):

PR #3046, opened 2021-10-28
fix(issuer-api): handle multiple certificates created at once

soanvig:
Fixes issue that some certificates were already stored in the system, but some were not


packages/traceability/issuer-api/src/pods/certificate/handlers/certificates-created.handler.ts
```diff
@@ -0,0 +1,129 @@
...
+        @InjectRepository(Certificate)
+        private readonly repository: Repository<Certificate>,
+        @InjectRepository(UnminedCommitment)
+        private readonly unminedCommitmentRepository: Repository<UnminedCommitment>,
+        private connection: Connection,
+        private readonly blockchainPropertiesService: BlockchainPropertiesService,
+        private readonly eventBus: EventBus
+    ) {}
+
+    async handle({ ids, privateInfo }: CertificatesCreatedEvent): Promise<void> {
```
soanvig, line 35: So event no longer returns ISuccessResponse, because... events' response cannot be handled anyway

packages/traceability/issuer-api/src/pods/certificate/handlers/certificates-created.handler.ts
```diff
@@ -0,0 +1,129 @@
...
+                id: In(ids)
+            },
+            relations: ['blockchain']
+        });
+
+        const notExistingCertificates = ids.filter(
+            (id) => !existingCertificates.some((c) => c.id === id)
+        );
+
+        existingCertificates.forEach((c) => {
```
soanvig, line 51: So it works like previously: if certificate is existing, Sync event is emitted, but only for existing certs

packages/traceability/issuer-api/src/pods/certificate/handlers/certificates-created.handler.ts
```diff
@@ -0,0 +1,129 @@
...
+            // Re-sync the certificate to sync any eventual changes
+            this.eventBus.publish(new SyncCertificateEvent(c.id));
+        });
+
+        if (notExistingCertificates.length === 0) {
+            return;
+        }
+
+        const newCertificates = await Promise.all(
+            notExistingCertificates.map(async (id) => {
```
soanvig, line 61: Therefore only "notExistingCertificates" are processed

packages/traceability/issuer-api/src/pods/certificate/handlers/certificates-created.handler.ts
```diff
@@ -0,0 +1,129 @@
...
+                try {
+                    return await new CertificateFacade(id, blockchainProperties.wrap()).sync();
+                } catch (e) {
+                    this.logger.error(e.message);
+                    throw e;
```
soanvig, line 66: The only difference is that if any of certificates in "batch" cannot be synced by issuer, then whole process fails, and no certificate is created
Not sure if that is acceptable

packages/traceability/issuer-api/src/pods/certificate/handlers/certificates-created.handler.ts
```diff
@@ -0,0 +1,129 @@
+import { EventBus, EventsHandler, IEventHandler } from '@nestjs/cqrs';
```
soanvig, line 1: Because of file rename it treats whole file as a new so I left some comments

soanvig: The  code is more-or-less the same

packages/traceability/issuer-api/src/pods/certificate/handlers/certificates-created.handler.ts
```diff
@@ -0,0 +1,129 @@
...
+        const txHash = newCertificates[0].creationTransactionHash;
+        const unminedCommitment = await this.checkCommitment(txHash);
+
+        const queryRunner = this.connection.createQueryRunner();
+        await queryRunner.connect();
+        await queryRunner.startTransaction();
+
+        try {
+            await Promise.all(
+                newCertificates.map(async (cert) => {
```
soanvig, line 88: All certificates are created in transaction

packages/traceability/issuer-api/src/pods/certificate/events/certificates-created-event.ts
```diff
@@ -3,10 +3,9 @@ interface IPrivateCertificateInfo {
     energy: string;
 }
 
-export class CertificateCreatedEvent {
+export class CertificatesCreatedEvent {
     constructor(
-        public readonly id: number,
-        public readonly byTxHash?: string,
+        public readonly ids: number[],
```
soanvig, line 8: This is the major change

packages/traceability/issuer-api/src/pods/certificate/handlers/certificates-created.handler.ts
josipbagaric: That's fine, it's better to fail early.

josipbagaric (approved):
👍

PR #3049, opened 2021-10-29
feat(admin-trades): Implementing API and UI

doga-foton:
**It's a work in progress.**

Related to:
OV-1262
OV-1261

API:

- [x] API endpoint has been added
- [ ] Query to get trades with relational data to list on UI

UI:

- [x] Menu item has been added
- [x] View has been added 

image

I would like to get feedback about;

- [x] If the implementation of the UI applied to guidelines.
- [ ] I needed to add two controllers on /trade/exchange and /trade/exchange-irec packages. 
- [ ] Existing trade.service has been used, and made getAll method as public, should I create a service only for this case?
- [ ] Folder / file structure


packages/ui/libs/user/logic/src/admin/trades.ts
```diff
@@ -0,0 +1,86 @@
...
+  deviceName: string;
+  volume: number;
+  price: number;
+  total: number;
+};
+
+const formatTrades: TFormatTrades = ({ trades, allDevices }) => {
+  const formattedTrades: FormattedTrade[] = [];
+
+  trades?.forEach((trade) => {
```
alexworker23, line 33: `const formattedTrades: FormattedTrade[] =  trades.map` will better fit here

packages/ui/libs/user/logic/src/admin/trades.ts
```diff
@@ -0,0 +1,86 @@
...
+  trades,
+  allDevices,
+}) => {
+  const { t } = useTranslation();
+  return {
+    header: {
+      date: t('admin.trades.date'),
+      buyer: t('admin.trades.buyer'),
+      seller: t('admin.trades.seller'),
+      energy: 'Energy',
```
alexworker23, line 76: please don't forget to localize this

packages/ui/libs/user/view/src/pages/AdminTradesPage/AdminTradesPage.effects.ts
```diff
@@ -0,0 +1,18 @@
+import { useFetchAllDevices } from '@energyweb/origin-ui-user-data';
+import { useApiAdminTrades } from '@energyweb/origin-ui-exchange-data';
```
alexworker23: Also pay attention to cross-package dependencies. 
I mean user related packages can use imports only from `user-${package}` or generic packages like `ui-core` `ui-utils` etc.
But we should avoid importing from other parts of the UI (certificate/exchange/device), since it will break their independency.

alexworker23 (commented):
In general: the implementation of the UI applies to guidelines ✅

packages/ui/libs/exchange/data/src/fetching/adminTrades.ts
```diff
@@ -0,0 +1,14 @@
+import {
+  TradeDTO,
+  useAdminTradeControllerGetAll,
+} from '@energyweb/exchange-irec-react-query-client';
+
+export const useApiAdminTrades = () => {
```
alexworker23: this hook should be located in `user/data` not `exchange/data`

PR #3051, opened 2021-10-29
chore(issuer): optimize cert sync, and events retrieval

soanvig:
I tried also to make getEvents accept more than one event type at once, but the impact on performance was almost not-relevant (and the solution was a bit complex and could make some issues), so I dropped it.

This solution looks like it mitigated the problem of increasing sync time the greater the ledger was.

Now I have almost constant re-sync time. I'm going to test it further (to see if time increases for more certificates) in a few days.

### DIFF
--- a/packages/traceability/issuer/src/blockchain-facade/Certificate.ts
+++ b/packages/traceability/issuer/src/blockchain-facade/Certificate.ts
@@ -199,9 +199,7 @@ export class Certificate implements ICertificate {
 
         const { registry } = this.blockchainProperties;
 
-        const issuanceTransaction = this.creationTransactionHash
-            ? await registry.provider.getTransactionReceipt(this.creationTransactionHash)
-            : await this.getIssuanceTransaction();
+        const issuanceTransaction = await this.getIssuanceTransaction();
 
         const certOnChain = await registry.getCertificate(this.id);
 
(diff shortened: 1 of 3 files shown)


packages/traceability/issuer/src/blockchain-facade/Certificate.ts
```diff
@@ -199,9 +199,7 @@ export class Certificate implements ICertificate {
 
         const { registry } = this.blockchainProperties;
 
-        const issuanceTransaction = this.creationTransactionHash
-            ? await registry.provider.getTransactionReceipt(this.creationTransactionHash)
-            : await this.getIssuanceTransaction();
+        const issuanceTransaction = await this.getIssuanceTransaction();
```
josipbagaric, line 202: Hmm what's the reason for this change?
This used to fix some bug, can't remember.

Check git blame.

soanvig: I changed `await this.getIssuanceTransaction();` as well, so now it contains this condition inside itself. I just simplified this, so I can reuse this "condition" everywhere and don't repeat myself.

josipbagaric: OK!

josipbagaric (approved):

PR #3062, opened 2021-11-02
feat(origin-organization-irec-api): support source trade acc for irec module

ioncreature:


packages/organizations/origin-organization-irec-api/src/irec/irec.service.ts
```diff
@@ -350,11 +351,12 @@ export class IrecService implements IIrecService {
         fromUser: UserIdentifier,
         toTradeAccount: string,
         assetId: string,
+        fromTradeAccount?: string,
         amount?: number
     ): Promise<TransactionResult> {
```
ioncreature, line 356: @JosephBagaric in case you wonder - I've checked nobody uses extended method signature, so it is safe to put the optional parameter in between

josipbagaric (approved):

alexworker23 (approved):

PR #3081, opened 2021-11-08
feat(issuer-api): allow to query certs by generationEnd frame

soanvig:
For 24/7 applications we can have more than 10k certificates. It would be actually really nice to have a way to query only for certificates from given time-span, instead for querying for all of them.


packages/traceability/issuer-api/package.json
```diff
@@ -25,7 +25,7 @@
         "test": "jest --env 'node'",
         "test:watch": "jest --watch",
         "test:cov": "jest --coverage",
-        "test:mocha": "mocha -r ts-node/register test/*.e2e-spec.ts --timeout 60000 --exit",
+        "test:mocha": "mocha -r ts-node/register test/certificate.e2e-spec.ts --timeout 60000 --exit",
```
josipbagaric: ☝️

josipbagaric:
@soanvig could you add some context on why this change is necessary?

packages/traceability/issuer-api/package.json
soanvig: done

josipbagaric:
Could we add some other filters as well?

soanvig:
I think we could but I'm not sure what may be useful

josipbagaric:
I think filters by `deviceId`, `generationStartTime` (similar to endtime), and `creationTime` would be useful.

soanvig:
> I think filters by `deviceId`, `generationStartTime` (similar to endtime), and `creationTime` would be useful.

done!

JonaszJestem (approved):

JonaszJestem (approved):

PR #3084, opened 2021-11-09
fix(issuer-api): fix certificate created transaction

soanvig:
Transaction was not working at all, at it probably caused the bug, that I queried for transaction which should contain certificates, that weren't ready in database yet. There is very small chance, that I was so lucky few times, but I have no other idea, why I received e.g. only 4 of 10 certs, that should be created in transaction (while later in database all were present).


josipbagaric (approved):
👌

PR #3085, opened 2021-11-09
feature(origin-backend) added password validation on user registration

Malak67:
Added password validation for user registration


josipbagaric (commented):
Hmm I don't see any tests.
Could you please add regression tests to cover the failing and passing use cases?

josipbagaric (approved):

PR #3118, opened 2021-11-15
fix(issuer-api): fix get all certs query

soanvig:
Certificates can have time fields greater than `new Date()`. For such fields and no date query specified no certificate was returned.
Actually I still use the same solution (just 2038 year instead of `new Date()`), because this is the most elegant way for building a query.


josipbagaric (approved):
👍

PR #3120, opened 2021-11-17
add API examples and descriptions for DTOs

wjpurdum:
Add examples and descriptions for Trade SDKs DTOs

### DIFF
--- a/packages/trade/exchange/src/pods/asset/asset.entity.ts
+++ b/packages/trade/exchange/src/pods/asset/asset.entity.ts
@@ -21,29 +21,31 @@ export class Asset extends ExtendedBaseEntity implements IAsset {
     @PrimaryGeneratedColumn('uuid')
     id: string;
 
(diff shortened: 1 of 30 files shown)


packages/trade/exchange/src/pods/asset/asset.entity.ts
```diff
@@ -21,29 +21,31 @@ export class Asset extends ExtendedBaseEntity implements IAsset {
...
     @IsNotEmpty()
     @IsString()
     address: string;
 
+    //Get Example
     @ApiProperty({ type: String })
     @Column()
     @IsNotEmpty()
     @IsString()
     tokenId: string;
```
josipbagaric: We can put

```js
@ApiProperty({
    type: String,
    description: "Token ID of the certificate that is being traded",
    example: "1" 
})
```

josipbagaric (approved):

PR #3144, opened 2021-11-22
feat/reset password ui

alexworker23:
Screenshot 2021-11-22 at 11 17 26
Screenshot 2021-11-22 at 11 17 55


packages/ui/libs/device/data/src/fetching/deviceFirstImage.ts
```diff
@@ -9,6 +9,8 @@ export const useDeviceFirstImageUrl = (
   const [imageUrl, setImageUrl] = useState('');
 
   const getAndSetImage = async (id: string) => {
+    if (id === 'null') return;
+
     const response = await publicFileDownloadHandler(id);
```
ioncreature: Looks a bit strange. Is it really `"null"` string? If yes, maybe it is better to fix somewhere before this happens
The same for `packages/ui/libs/device/data/src/fetching/deviceImages.ts:11`

alexworker23: I agree, this is how i received it from `/api/device-registry` GET:
Screenshot 2021-11-25 at 09 59 13

ioncreature: So this is definitely a bug on backend, I'll check what is going on there

Malak67:
Is there a key/token which will be added in the URL received in the email which then redirects to Change password?
I'm just wondering because I did not see a way to identify the account which is changing the password. But maybe you're waiting for the backend to be done.

alexworker23:
> Is there a key/token which will be added in the URL received in the email which then redirects to Change password? I'm just wondering because I did not see a way to identify the account which is changing the password. But maybe you're waiting for the backend to be done.

Yes, there will be a token, but the backend implementation for this is not done yet. That's why I left submitHandlers as empty functions with comments above them.

packages/ui/libs/device/data/src/fetching/deviceFirstImage.ts
ioncreature: it was a bug on the backend, added a task to validate imageIds.

ioncreature (approved):

Malak67 (approved):

driamend (approved):

PR #3174, opened 2021-11-29
feat/exchange claim UI

alexworker23:
Add Claim action to Exchange Inbox:
Screenshot 2021-12-07 at 08 16 11

Add certificates claimed from the Exchange to the Claims Report table:
Screenshot 2021-12-07 at 08 17 01

Adjust the Certificate Detail View to display claim-related data for Exchange claimed certificates:
Screenshot 2021-12-07 at 08 23 48


josipbagaric (approved):
Backend side looks good.

ioncreature (approved):

alexworker23:
> Backend side looks good.

Only thing I noticed is Claimed Energy displayed as 0. 
This happens because, `irec/certificate/:id` returns 
`energy: {publicVolume: "0", privateVolume: "0", claimedVolume: "0"}`. 
So I guess claims from the exchange are not reflected in the energy volumes, though they are received in `claims` property in the same response.

josipbagaric:
> > Backend side looks good.
> 
> Only thing I noticed is Claimed Energy displayed as 0. This happens because, `irec/certificate/:id` returns `energy: {publicVolume: "0", privateVolume: "0", claimedVolume: "0"}`. So I guess claims from the exchange are not reflected in the energy volumes, though they are received in `claims` property in the same response.

Yes, that's actually expected, because claiming from the exchange doesn't claim to the user's blockchain address (it has to support cases where the user doesn't want to be the custodian). Instead, it claims it to the platform address, but stores the user ID as the beneficiary.

PR #3197, opened 2021-12-07
feat/blockchain account options

alexworker23:
Adjust the UI to allow user have 3 different flows with certificates:
- [x] Custody NO org blockchain address
- [x] Custody WITH org blockchain address
- [x] Full self ownership
Screenshot 2021-12-08 at 09 05 05


packages/trade/exchange/src/pods/transfer/transfer.entity.ts
```diff
@@ -61,4 +61,9 @@ export class Transfer extends ExtendedBaseEntity {
     @Column('simple-json', { nullable: true })
     @IsOptional()
     claimData?: IClaimData;
+
+    @ApiProperty({ type: String, required: false })
+    @Column({ nullable: true })
+    @IsOptional()
+    to?: string;
```
alexworker23: @JosephBagaric should I add this column or I am supposed to use the `address` instead?

josipbagaric: Good question. You should use the `address` field instead.

alexworker23: Unfortunately I was unable to do so, because in `address` we store the user's exchange deposit address and it is always truthy. 
Changing `to` to `address` also broke the claiming from the exchange for users without organization blockchain address.
So I am thinking of simply renaming the columns to match their purpose:
`address` -> `exchangeAddress`
`to` -> `blockchainAddress`
What do you think @JosephBagaric ?

alexworker23: well, changing `address` to `exchangeAddress` can also unexpectedly brake several things. 
so I will just rename `to` to `claimAddress` and that's it.

josipbagaric (approved):

PR #3203, opened 2021-12-10
Feat/add account code select

driamend:
- added irecTradeAccountCode select field to Device creation & Certificate Request creation
- created SINGLE_ACCOUNT_MODE env variable and passed it to the DeviceApp
- added SINGLE_ACCOUNT_MODE env to deploy and release config
- added description for SINGLE_ACCOUNT_MODE env variable

chrome_daB6nsxCC5
chrome_GDoAE68qKi


packages/ui/libs/device/data/src/fetching/myAccounts.ts
```diff
@@ -0,0 +1,25 @@
...
+  getConnectionControllerGetMyAccountsQueryKey,
+  useConnectionControllerGetMyAccounts,
+} from '@energyweb/origin-organization-irec-api-react-query-client';
+import { useQueryClient } from 'react-query';
+
+export const useApiMyAccounts = () => {
+  const queryClient = useQueryClient();
+  const key = getConnectionControllerGetMyAccountsQueryKey();
+
+  const isSingleAccountMode = /true/i.test(process.env.NX_SINGLE_ACCOUNT_MODE);
```
alexworker23: Check how the env variables are supplied at the `origin-ui` level

packages/ui/libs/device/data/src/utils.ts
```diff
@@ -89,6 +89,7 @@ export function decomposeForIRec({
       ? newDevice.timeZone[0].value.toString()
       : timeZones[0].timeZone,
     gridOperator: newDevice.gridOperator,
+    irecTradeAccountCode: newDevice?.irecTradeAccountCode,
```
alexworker23: should be 
`irecTradeAccountCode: newDevice?.irecTradeAccountCode || undefined`
to ensure that if the irecTradeAccountCode is not allowed to input for the user there will be no initialValues sent ('')

packages/ui/libs/device/view/src/pages/RegisterPage/RegisterPage.effects.ts
```diff
@@ -27,6 +28,7 @@ export const useRegisterPageEffects = () => {
     useAllDeviceFuelTypes();
   const { allTypes: allDeviceTypes, isLoading: areDeviceTypesLoading } =
     useAllDeviceTypes();
+  const { myAccounts, isLoading: areMyAccountsLoading } = useApiMyAccounts();
```
alexworker23: you should also check if there are cases when this endpoint is forbidden (e.g. user without organization, user with inactive organization, user without irec org) and add `enabled` prop. 
Also it should be disabled if `SINGLE_ACCOUNT_MODE` is false

alexworker23 (approved):
LGTM 👍

PR #3226, opened 2021-12-16
feat/signing adapter

soanvig:
- feat(issuer-api): allow to set signingAdapter in issuer-api configuration refactor: unify usage of blockchain properties wrap feat: use custom adapter for blockchain properties refactor(issuer-api): add options module
- feat: encrypt operator platform key BREAKING CHANGE: - user needs to provide either encryption key or custom signer adapter in IssuerModule config - existing operator keys are automatically migrated and encrypted (during first usage) - it's impossible to use BlockchainPropertiesModule without IssuerApi module - new entity (SignerEntity) needs to be added to TypeORM.forRoot - if blockchainProperties are created manually via SQL, then `issuer_signer` table needs to be populated with encryption key, that can be created using `encryptOperatorKey` function - if user cannot use whole IssuerModule, but is using BlockchainPropertiesModule, then OptionsModule needs to be registered at least
- fix: add encryption key to origin applications

OPERATOR_ENCRYPTION_KEY needs to be configured on environments


.env.example
```diff
@@ -34,6 +34,7 @@ SUPPORTED_NETWORK_IDS=1337;73799;246
 
 # Secret keys
 DEPLOY_KEY=d9066ff9f753a1898709b568119055660a77d9aae4d7a4ad677b8fb3d2a571e5
+OPERATOR_ENCRYPTION_KEY=somerandompassword
```
josipbagaric, line 39: This should be marked as optional, since the users of the SDK might use their own adapter.

This variable would then be used only as a fallback. Let's make this clear in .env.example

packages/traceability/issuer-api/src/pods/options/types.ts
```diff
@@ -0,0 +1,21 @@
+import { Signer } from 'ethers';
+
+export type SignerAdapter = (blockchainNetId: number) => Promise<Signer>;
+
+export interface IssuerModuleOptions {
+    enableTransactionLogging: boolean;
+    enableCertificationRequest: boolean;
+    getOperatorEncryptionKey?: () => string;
```
josipbagaric: Why not:
1. Create a `EncryptedPrivateKeyAdapter`
2. Require `EncryptedPrivateKeyAdapter` to have the encryption key set up
2. Rename `customSignerAdapter` to `signerAdapter`
3. Set `EncryptedPrivateKeyAdapter` as the default `signerAdapter`?

Having these two issuer module optinos is redundant, since they relate to the same functionality.

soanvig: Well, yeah, more or less that would be a good idea.
I had some problem with this approach initially, but I now just figured out how to resolve this.

josipbagaric: IssuerModule shouldn't really care about the encryption key, only the EncryptedPrivateKeyAdapter  should.

.env.example
soanvig: Done, added comment

soanvig:
@JosephBagaric for practical reasons it works like that:

By default EncrypterSignerAdapter is loaded. It uses process.env.OPERATOR_ENCRYPTION_KEY implicitly.
If you provided SignerAdapter provider anywhere in the application (literally anywhere) - it will be used.
This way as signer creator you can utilize Nest.js injection without really caring how this adapter should be forwarded to issuer-api module (which makes sense especially from 247-sdk which abstractes issuer-api totally away).

Why it's done this way - because it's the easiest way to keep it scoped (in terms of module) and you can really easy use it in any already built application (the injection is easy, and you can do anything).

It is described in README and breaking change note. Migration to encrypted signer is automatic

It works better now, I have to agree. At least it's really easy to use :)

packages/traceability/issuer-api/migrations/1639660016195-AddSignerTable.ts
```diff
@@ -0,0 +1,49 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddSignerTable1639660016195 implements MigrationInterface {
+    name = 'AddSignerTable1639660016195';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            `CREATE TABLE "issuer_signer" ("blockchainNetId" integer NOT NULL, "platformOperatorPrivateKey" character varying NOT NULL, "isEncrypted" boolean NOT NULL, CONSTRAINT "PK_7c8b739645fb8518b866b45a8bb" PRIMARY KEY ("blockchainNetId"))`
```
josipbagaric, line 8: Why is the blockchainNetId needed for the signer?

Signer can sign messages/transactions without ever communicating with the blockchain.

Also, we don't really support multiple networks on the Issuer API level, so we will not have different signers for different networks.

If we force this table to only have a single row, and always use that signer, that should be enough for our use case.

packages/traceability/issuer-irec-api/test/certificate.e2e-spec.ts
```diff
@@ -108,7 +108,7 @@ describe('Certificate tests', () => {
         await app.close();
     });
 
-    it('should claim a certificate', async () => {
+    it.only('should claim a certificate', async () => {
```
josipbagaric: Only.

josipbagaric (changes requested):

packages/traceability/issuer-api/migrations/1639660016195-AddSignerTable.ts
soanvig: Ok, I will simplify this :-)

soanvig: My intention was different signer for different network, but yeah, since we don't support this, then I guess it's whatever.

soanvig: But... After all @JosephBagaric  I wonder if this is such a problem.
Because this is possible to create multiple blockchain properties (and it is, because this is not restricted anywhere, despite that it's impossible to use different blockchain properties in practice), why just don't keep this relation?
For me at least this would be more intuitive.

josipbagaric: @soanvig You're right, I thought it was restricted. Let's keep it then.

packages/traceability/issuer-irec-api/test/certificate.e2e-spec.ts
soanvig: done

josipbagaric (approved):

PR #3238, opened 2021-12-22
fix/organization app for unlogged

alexworker23:
Fixed issue with Organization UI app when unlogged user gets on those routes - it will start re-rendering infinitely and trigger infinite requests for `/user/me`

### DIFF
--- a/packages/ui/apps/origin-ui/src/hooks/useInterceptors.ts
+++ b/packages/ui/apps/origin-ui/src/hooks/useInterceptors.ts
@@ -21,12 +21,14 @@ export const useAxiosDefaults = () => {
   useEffect(() => {
     const token = getAuthenticationToken();
     axios.defaults.baseURL = window.config.BACKEND_URL;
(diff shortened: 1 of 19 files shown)


packages/ui/apps/origin-ui/src/hooks/useInterceptors.ts
```diff
@@ -21,12 +21,14 @@ export const useAxiosDefaults = () => {
   useEffect(() => {
     const token = getAuthenticationToken();
     axios.defaults.baseURL = window.config.BACKEND_URL;
-    axios.defaults.headers.common['Authorization'] = `Bearer ${token}`;
+    if (!!token) {
+      axios.defaults.headers.common['Authorization'] = `Bearer ${token}`;
+    }
```
ioncreature: probably ```if (token) {``` would be a bit easier to read

ioncreature (approved):

PR #3239, opened 2021-12-23
fix(origin-ui) Fix smart meter reads and allowing certificates bulk t…

Malak67:
Fixing smart meter reads
Allowing bulk transfer and claim of certificates

### DIFF
--- a/packages/ui/libs/certificate/view/src/containers/list/CertificateActionContent/CertificateActionContent.effects.ts
+++ b/packages/ui/libs/certificate/view/src/containers/list/CertificateActionContent/CertificateActionContent.effects.ts
@@ -24,6 +24,7 @@ export const useCertificateActionContentEffects = <Id>(
     items: CertificateActionContentProps<Id>['selectedItems']
   ) => {
     if (ids.length > energyAmounts.length) {
+      let newItems: EnergyAmounts<Id>[] = [];
       ids.forEach((id) => {
         const alreadyIncluded = energyAmounts.find(
           (amount) => amount.id === id
@@ -36,9 +37,10 @@ export const useCertificateActionContentEffects = <Id>(
               items.find((item) => item.id === id).energy.replace(/,/g, '')
             ).toString(),
           };
-          setEnergyAmounts([...energyAmounts, preparedNewItem]);
+          newItems.push(preparedNewItem);
         }
       });
+      setEnergyAmounts([...energyAmounts, ...newItems]);
     }
     if (ids.length < energyAmounts.length) {
       energyAmounts.forEach((amount) => {
--- a/packages/ui/libs/certificate/view/src/containers/list/CertificateActionContent/CertificateActionContent.tsx
+++ b/packages/ui/libs/certificate/view/src/containers/list/CertificateActionContent/CertificateActionContent.tsx
@@ -21,6 +21,7 @@ export interface CertificateActionContentProps<Id> {
   >[];
   buttonDisabled?: boolean;
   setTotalAmount?: (newValue: number) => void;
+  disableBulkActions?: boolean;
 }
 
 export type TCertificateActionContent = <Id>(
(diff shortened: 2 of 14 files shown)


packages/ui/libs/certificate/view/src/containers/list/CertificateActionContent/CertificateActionContent.effects.ts
```diff
@@ -36,9 +37,10 @@ export const useCertificateActionContentEffects = <Id>(
               items.find((item) => item.id === id).energy.replace(/,/g, '')
             ).toString(),
           };
-          setEnergyAmounts([...energyAmounts, preparedNewItem]);
+          newItems.push(preparedNewItem);
         }
       });
+      setEnergyAmounts([...energyAmounts, ...newItems]);
```
alexworker23, line 43: why do you change this logic? did you encounter any bug related to selection of items and this is the fix?

packages/ui/libs/certificate/view/src/containers/list/CertificateActionContent/CertificateActionContent.tsx
```diff
@@ -107,13 +100,6 @@ export const CertificateActionContent: TCertificateActionContent = ({
           </>
         )}
       </Button>
-      {selectedIds.length > 1 && (
```
alexworker23, line 110: Have you checked if the bulk actions work with all the available actions of both Exchange and Blockchain inboxes ?

alexworker23 (changes requested):
`index.html` is auto-generated file and should not be included

alexworker23:
Also please follow our conventional-commit rules and commit changes per package to keep our changelogs updated accordingly

packages/ui/libs/certificate/view/src/containers/list/CertificateActionContent/CertificateActionContent.effects.ts
Malak67: Yes. When selecting multiple at once, only the value for the last one is taken into account. This was the fix for it.

packages/ui/libs/certificate/view/src/containers/list/CertificateActionContent/CertificateActionContent.tsx
Malak67: I have tested this for Blockchain inbox (locally with D-REC) cannot test in Origin. Need your help here.

alexworker23: Ok, will check it

Malak67:
> Also please follow our conventional-commit rules and commit changes per package to keep our changelogs updated accordingly

Can you give me an example on how to rename the commits?

alexworker23:
> Can you give me an example on how to rename the commits?

#3235 for example

Malak67:
> `index.html` is auto-generated file and should not be included

Initially I did not add it but I noticed it failed the Ling & Test because of it. I will remove it again.

alexworker23:
Here is how it looks like:
- Exchange Inbox
1. Sell - not working as bulk actions only posts the first one for sale
2. Withdraw - doesn't work at all with multiple certificates
3. Transfer - behaves weirdly. Transfers the first certificate to user's blockhain address (behaves as withdrawal) and the last one to the actual address mentioned in inputs. So we should probably disable bulk operations for this action for now.
4. Export - unable to properly test, but for the safety of operations lets also assume it couldn't handle multiple 
5. Claim - claims only the first one, for the second one errors. 

- Blockchain Inbox
1. Deposit - works fine
2. Retire - works fine 
3. Transfer - works fine

So I would say that we should allow bulk actions only for Blockchain Inbox

alexworker23:
You have to have 3 commits min:
1 - certificate
2 - device
3 - localization

alexworker23 (approved):

PR #3270, opened 2021-12-28
feat: certificate id is bigint

ioncreature:
BREAKING CHANGE: certificate and certification requests id's are big ints now


alexworker23 (commented):
I think we have to adjust this on the UI level as well?

ioncreature:
> I think we have to adjust this on the UI level as well?

Yep, we are

PR #3298, opened 2022-01-14
Remove transaction log from issuer-api

JonaszJestem:


soanvig (changes requested):
What about migration dropping database?

soanvig (approved):

PR #3300, opened 2022-01-17
Chore/add UI core tests forms tables

driamend:


JonaszJestem (approved):
I'm not an expert in frontend unit testing. Selecting elements by specific material ui classes seems to be volatile to any material library update. Everything else looks good

soanvig:
> I'm not an expert in frontend unit testing. Selecting elements by specific material ui classes seems to be volatile to any material library update. Everything else looks good

It's better to select them by dedicated attribute, like `data-test-id`

PR #3302, opened 2022-01-20
chore(origin-ui-core): add unit tests for lists and cards

driamend:


soanvig (approved):
Besides selecting elements by material classes instead of dedicated `data-test-id` (or any other attribute) it's ok

packages/ui/libs/ui/core/src/__tests__/Lists & Cards/GenericItemsList.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <Checkboxes {...(Checkboxes.args as GenericItemsListProps<any, any>)} />
+    );
+
+    const itemsCount =
+      Checkboxes.args.listContainers.reduce(
+        (acc, current) => acc + current.containerItems.length,
+        0
+      ) +
+      Checkboxes.args.listContainers.length +
+      1;
```
JonaszJestem: some magic values. I guess that can be extracted:
```
const itemsInLists = Checkboxes.args.listContainers.reduce(
        (acc, current) => acc + current.containerItems.length,
        0
      )
```
And why is there `+1` at the end?

packages/ui/libs/ui/core/src/__tests__/Lists & Cards/GenericItemsList.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
```
JonaszJestem: some magic values. I guess that can be extracted:
```
const itemsInLists = Checkboxes.args.listContainers.reduce(
        (acc, current) => acc + current.containerItems.length,
        0
      )
```
And why is there `+1` at the end?

JonaszJestem (approved):

PR #3325, opened 2022-01-31
feat: allow issuer to create irec connection

soanvig:


packages/organizations/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
```diff
@@ -85,48 +85,31 @@ export class CreateConnectionHandler implements ICommandHandler<CreateConnection
...
     ) {
-        const platformAdmin = await this.userService.getPlatformAdmin();
         const organization = await this.organizationService.findOne(organizationId);
         const irecOrganization = await this.irecService.getUserOrganizationByTokens(
             clientId,
             clientSecret,
             tokens
         );
 
-        // check the IREC account against the platform organization's IREC account requirements
```
JonaszJestem, line 96: What happened to that check? Isn't it needed now?

JonaszJestem (approved):

packages/organizations/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
soanvig: The logic is the same, but it's more explicit now (previously it was negative condition, not I have positive conditions and if they are not met, it throws this error)

PR #3331, opened 2022-02-07
feat: allow generic claim data

soanvig:
Reason: We want claim data to be generic, because enforcing specific fields does not apply to all applications

Solution:
1. Make claim data to be generic JSON-serializable fields
2. Add schema version to issuer - parsing claim data from contract will require this to understand, how it should be parsed (either backward compatible or not)
3. Adjust code in all origin applications, that import claim data interface from issuer

### DIFF
--- a/packages/organizations/origin-organization-irec-api/src/irec/irec.service.ts
+++ b/packages/organizations/origin-organization-irec-api/src/irec/irec.service.ts
@@ -46,14 +46,16 @@ export interface ICreateBeneficiary {
     location: string;
 }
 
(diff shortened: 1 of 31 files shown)


soanvig:
@kosecki123 please take a look - this is a rough idea of an implementation. WDYT?

kosecki123 (approved):

kosecki123:
> @kosecki123 please take a look - this is a rough idea of an implementation. WDYT?

Looks good. 

Migration script should go through but we will get the result after deployment to canary.

PR #3334, opened 2022-02-10
fix: fix irec account creation

soanvig:


packages/organizations/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
```diff
@@ -101,14 +101,16 @@ export class CreateConnectionHandler implements ICommandHandler<CreateConnection
...
             ]);
         } else {
-            throw new BadRequestException(
-                'IREC account organization has to have issuer or registrant or participant role'
-            );
+            if (!isRegistrant) {
+                throw new BadRequestException(
+                    'IREC account organization has to have issuer or registrant or participant role'
+                );
+            }
```
JonaszJestem, line 113: so being registrant doesn't create the accounts?

JonaszJestem (approved):

packages/organizations/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
soanvig: Exactly, old implementation was not creating that as well, I refactored it in wrong way. Now it is correct

soanvig: Registrant cannot have trade account created

PR #3337, opened 2022-02-22
feat(exchange): return all trades with additional data for admin

soanvig:
Issue: https://energyweb.atlassian.net/browse/OV-1261

So the idea is that Admin now will be able to see all trades (not his own only), and will receive more information about the trades as well (asset.tokenId, and seller/buyer ids)


soanvig:
@driamend please take a look if new DTO (TradeForAdminDTO) suits frontend case

JonaszJestem (approved):

PR #3340, opened 2022-02-25
Feat/admin trades page

driamend:
- Added admin trades page

Selection_256

### DIFF
--- a/packages/ui/libs/user/logic/src/admin/trades.ts
+++ b/packages/ui/libs/user/logic/src/admin/trades.ts
@@ -1,0 +1,72 @@
+import { TableComponentProps, TableRowData } from '@energyweb/origin-ui-core';
+import { TradeForAdminDTO } from '@energyweb/exchange-irec-react-query-client';
+import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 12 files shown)


packages/ui/libs/user/logic/src/admin/trades.ts
```diff
@@ -0,0 +1,72 @@
...
+  price: number;
+  total: string;
+};
+
+const formatTrades: TFormatTrades = ({ trades }): FormattedTrades[] => {
+  return (
+    trades?.map((trade) => {
+      const price = trade.price / 100;
+      const totalPrice =
+        parseFloat(EnergyFormatter.format(trade.volume)) * price;
```
JonaszJestem, line 35: I'm not sure what numbers are there exactly, but to minimize losing precision it would be better to multiply it first and then divide by `100`. Also I'm not sure what `100` is? Is that `centsToUSD` ratio?

JonaszJestem (approved):

packages/ui/libs/user/logic/src/admin/trades.ts
driamend: it's an implementation from other pages so I didn't change it

PR #3344, opened 2022-03-17
fix(issuer): Increment ID when batch issuing

josipbagaric:
Screen Shot 2022-03-17 at 17 13 45
.

### DIFF
--- a/packages/traceability/issuer/contracts/Registry.sol
+++ b/packages/traceability/issuer/contracts/Registry.sol
@@ -69,6 +69,8 @@ contract Registry is ERC1155, ERC1888 {
 				data: _data[i]
 			});
 		}
+		
+		_latestCertificateId = ids[ids.length - 1];
 
 		emit IssuanceBatch(operator, _topics, ids, _values);
 	}


JonaszJestem (approved):

soanvig (approved):
It's interesting that we haven't encountered this problem in any practical matter.

I will take over test, so wait for commit @josipbagaric  :)

dwojno (approved):

soanvig:
OK I know why we haven't encountered that. So we use `issueBatch` from `Issuer` contract which does everything through certification request and approval. If we used `issueBatch` from `Registry`, then we would have a problem

josipbagaric:
> OK I know why we haven't encountered that. So we use `issueBatch` from `Issuer` contract which does everything through certification request and approval. If we used `issueBatch` from `Registry`, then we would have a problem

`issueBatch()` from the Issuer.sol uses `issueBatchMultiple()` from RegistryExtended.sol instead of `issueBatch()`. That's why it's never used.

josipbagaric:
Feel free to add a test to this branch and merge this whenever.

PR #3348, opened 2022-04-20
fix: IREC integration issues

soanvig:
- fix(ui): fix buy validation in buyDirect
- fix(irec): partial claim correct amount instead of all available
- fix(irec): partial export correct amount instead of all available
- fix(irec): generate unique account code
- fix(ui): send certification request time without timezone

### DIFF
--- a/packages/organizations/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
+++ b/packages/organizations/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
@@ -3,6 +3,7 @@ import { CommandHandler, EventBus, ICommandHandler } from '@nestjs/cqrs';
 import { InjectRepository } from '@nestjs/typeorm';
 import { BadRequestException, Inject } from '@nestjs/common';
 import { Organization, OrganizationService } from '@energyweb/origin-backend';
+import { randomBytes } from 'crypto';
 
 import { RegistrationService } from '../../registration';
 import { IREC_SERVICE, IrecService } from '../../irec';
@@ -129,8 +130,11 @@ export class CreateConnectionHandler implements ICommandHandler<CreateConnection
 
         for (const account of accounts) {
             if (!irecAccounts.some((a) => a.type === account)) {
+                const numberOfChars = 2;
+                const randomCode = randomBytes(numberOfChars * 2).toString('hex'); // 1 byte is 2 characters
+
                 await this.irecService.createAccountByTokens(clientId, clientSecret, tokens, {
-                    code: `${account}-account`,
+                    code: `${account}-${randomCode}`,
                     type: account,
                     name: `origin ${account} account`,
                     private: false,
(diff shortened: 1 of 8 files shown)


JonaszJestem (approved):

packages/organizations/origin-organization-irec-api/src/irec/irec.service.ts
```diff
@@ -420,7 +424,9 @@ export class IrecService implements IIrecService {
 
         const transferItem = new ReservationItem();
         transferItem.code = item.code;
-        transferItem.amount = item.volume;
+
+        // Optionally convert Wh to MWh
+        transferItem.amount = amount ? Number(amount) / 1_000_000 : item.volume;
```
JonaszJestem, line 429: sometimes `amount` is in Wh and sometimes in MWh, wouldnt't that be confusing?

packages/organizations/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
```diff
@@ -129,8 +130,11 @@ export class CreateConnectionHandler implements ICommandHandler<CreateConnection
 
         for (const account of accounts) {
             if (!irecAccounts.some((a) => a.type === account)) {
+                const numberOfChars = 4;
+                const randomCode = randomBytes(numberOfChars / 2).toString('hex'); // 1 byte is 2 characters
```
JonaszJestem, line 134: isn't that the other way around? 1 byte -> 8 bits, hex needs 16bits so 2 bytes is 1 character. But the implementation is ok

packages/organizations/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
```diff
@@ -129,8 +130,11 @@ export class CreateConnectionHandler implements ICommandHandler<CreateConnection
...
+
                 await this.irecService.createAccountByTokens(clientId, clientSecret, tokens, {
-                    code: `${account}-account`,
+                    code: `${account}-${randomCode}`,
```
JonaszJestem, line 137: it would probably not be unique at some point, is that correct?

JonaszJestem (approved):

packages/organizations/origin-organization-irec-api/src/irec/irec.service.ts
soanvig: It's in wh in this case, and the parameter is documented

packages/organizations/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
soanvig: No, in that case it is 1 byte = 2 chars.

obraz

Not 100% sure why

patryk-kochanski (approved):

packages/organizations/origin-organization-irec-api/src/connection/handlers/create-connection.handler.ts
soanvig: Well, the original problem was that it was not unique. Now it will be. We have 16 characters in hex, which gives us 65k combinations. Should be more than sufficient. If it's not, then we can always increase this.

If account is not unique it will be rejected by irec. In such case you can just retry to get another combination, and if this happens too often, then devs can just increase number of random chars.

So in IREC response i got info that it would be nice if account code is easy to copy or write manually, that's why I went with 4 random chars only.

PR #3367, opened 2022-10-24
Feat/gp 243/set transaction fees at deploy

knzeng-e:
This PR add the ability to set fee parameters on contract contract deployment

### DIFF
--- a/packages/origin-backend/package.json
+++ b/packages/origin-backend/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@energyweb/origin-backend",
-    "version": "11.2.1",
+    "version": "11.2.2",
     "description": "API for managing users, certificate, requests and devices in the Origin project",
     "main": "dist/js/src/index.js",
     "scripts": {
--- a/packages/origin-backend/package.json
+++ b/packages/origin-backend/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@energyweb/origin-backend",
-    "version": "11.2.2",
+    "version": "11.2.1",
     "description": "API for managing users, certificate, requests and devices in the Origin project",
     "main": "dist/js/src/index.js",
     "scripts": {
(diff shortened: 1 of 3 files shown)


packages/traceability/issuer/src/migrate.ts
```diff
@@ -7,15 +7,24 @@ import { Issuer } from './ethers/Issuer';
 import { RegistryExtended } from './ethers/RegistryExtended';
 import { PrivateIssuer } from './ethers/PrivateIssuer';
 
+type DeployParameters = {
```
JonaszJestem: it would be better if `DeployParameters` will be exported from the origin issuer package so it can be used in other packages without redefining it.

packages/origin-backend/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@energyweb/origin-backend",
-    "version": "11.2.1",
+    "version": "11.2.2",
```
JonaszJestem: why do you want to bump that version?

packages/traceability/issuer/src/migrate.ts
knzeng-e: Do we have a dedicated "Issuer" package in origin ?

knzeng-e: is it the "packages/traceability/issuer" one ?

packages/origin-backend/package.json
knzeng-e: misunderstanding. I wil revert

knzeng-e: reverted on commit [8cf045a](https://github.com/energywebfoundation/origin/pull/3367/commits/8cf045a7655a174211f38dd1f121efe8d0f7c0bc)

packages/traceability/issuer/src/migrate.ts
knzeng-e: I exported it from constants.ts under the issuer package. Could you check commit [8cf045a](https://github.com/energywebfoundation/origin/pull/3367/commits/8cf045a7655a174211f38dd1f121efe8d0f7c0bc) please ?

JonaszJestem (approved):

